PR #241, opened 2022-08-19
fix: update the sample github repos

xiaofei-du:
Because

- the sample GitHub repos in the field description are non-exist

This commit

- fix the GitHub repos


EiffelFly (approved):
LGTM !

PR #312, opened 2022-12-22
chore: rename Visual Data Preparation to Versatile Data Pipeline

xiaofei-du:
Because

- instead of Visual Data Preparation, VDP is Versatile Data Pipeline now. We realise that as a general ETL infrastructure, VDP is capable of processing all kinds of unstructured data, and we should not limit its usage to only visual data. That's why we replace the word Visual with Versatile. Besides, the term Data Preparation is a bit misleading, users often think it has something to do with data labelling or cleaning. The term Data Pipeline is definitely more precise to capture the core concept of VDP.
 
This commit

- update to Versatile Data Pipeline

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-Visual Data Preparation (VDP) console
+Versatile Data Pipeline (VDP) console
 
 # About VDP Console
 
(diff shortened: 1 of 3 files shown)


EiffelFly (approved):
LGTM!

PR #371, opened 2023-03-14
feat: refactor <NameCell /> for the resource list to be anchor 

iamnamananand996:
will close this issue https://github.com/instill-ai/console/issues/368

Because

- Currently now the item in the resource list are displayed as button so user can't select and copy the value from it.
[screen-capture.webm](https://user-images.githubusercontent.com/31537362/225113315-ea1a0299-4d95-473b-97d9-c0be03b5a4fa.webm)

This commit

- made the item as anchor and it will be selectable
[screen-capture (1).webm](https://user-images.githubusercontent.com/31537362/225113275-4ede9314-a995-4bc3-be19-dcc48d2aa4af.webm)

### DIFF
--- a/src/components/ui/TableCells/NameCell.tsx
+++ b/src/components/ui/TableCells/NameCell.tsx
@@ -17,6 +17,10 @@ export type NameCellProps = {
   displayStateIndicator: boolean;
 } & CellBaseProps;
 
+export type CellItemProps = {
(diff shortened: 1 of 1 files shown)


iamnamananand996:
Hi @EiffelFly, this is hanging here from long time. 
Please review it once, either the changes, I have made is correct according to the opened issue or not.

Issue #407, opened 2023-04-22
enhancement: onboarding form validation


iamnamananand996:
onboarding form start validating all the field, when user just have add value for 1st field and it start showing errors for all fields, before adding any value.

the validation should be made only after user have clicked on start button at least once.

![Screenshot 2023-04-22 130140](https://user-images.githubusercontent.com/31537362/233770051-9e3562bc-421a-4acd-a534-6cbbe24eeea7.png)

cc - @EiffelFly

iamnamananand996:
this is just minor UX enhancement, I can fix this.

EiffelFly:
Thanks for reporting! And yes! feel free to pick up this issue

EiffelFly:
@iamnamananand996 You can check #409, I go further to separate the validation into two part. 

- Per-field validation
- Submit validation

Which will achieve what you proposed in this issue

iamnamananand996:
Hi @EiffelFly , I think you have read my mind 😄, it is exactly same what I was expecting. Thank for the implementation.

PR #408, opened 2023-04-22
fix: onboarding form validation

iamnamananand996:
will close https://github.com/instill-ai/console/issues/407

Because

- onboarding form start validating all the field, when user just have add value for 1st field and it start showing errors for all fields, before adding any value.

This commit

- have validation made only after user have clicked on start button at least once.

I have added some extra UX like : (`if field have error will auto focus that input field`), we can also have scroll to that particular field, that becomes very useful if we have longer forms.

### DIFF
--- a/src/components/OnboardingForm.tsx
+++ b/src/components/OnboardingForm.tsx
@@ -66,7 +66,6 @@ export const OnboardingForm = () => {
 
   const handleFieldChange = useCallback(
     (key: keyof OnboardingFormValues, event: ChangeEvent<HTMLInputElement>) => {
-      setFormIsDirty(true);
       setFieldValues((prev) => ({
         ...prev,
         [key]:
@@ -81,7 +80,6 @@ export const OnboardingForm = () => {
   const handleRoleChange = useCallback(
     (option: Nullable<SingleSelectOption>) => {
       if (!option) return;
-      setFormIsDirty(true);
       setSelectedRoleOption(option);
       setFieldValues((prev) => ({
         ...prev,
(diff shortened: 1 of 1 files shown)


src/components/OnboardingForm.tsx
```diff
@@ -126,12 +122,37 @@ export const OnboardingForm = () => {
...
+        var email: HTMLElement | void = document
+          .getElementById("email")
+          ?.focus();
+        return false;
+      }
+      if (errors.orgName) {
+        var orgName: HTMLElement | void = document
+          .getElementById("org-name")
+          ?.focus();
+        return false;
```
EiffelFly, line 139: I think you don't need to catch this variable, Even you need it, it's better to use let

src/components/OnboardingForm.tsx
```diff
@@ -126,12 +122,37 @@ export const OnboardingForm = () => {
...
+      }
     }
+    return false;
+  };
+
+  useEffect(() => {
+    validateForm();
   }, [fieldValues, formIsDirty]);
```
EiffelFly, line 147: I am confused, you call this function but it will immediately return so i will assume that you don't need to call this function at all

EiffelFly (commented):
@iamnamananand996 Our ci/cd is kind of complicated right now, so contributors can not run our ci/cd. Before we come up with a better ci/cd workflow we can't merge this PR without test. So I will create another PR to fix this issue and add you as co-author to compensate.

When it comes to the solution of the issue, let don't overthink this issue. Just validate the form when user submit it. 

Besides that, there are multiple weird code in your PR, please check it as reference in the future contribution, Thanks for your help!

src/components/OnboardingForm.tsx
iamnamananand996: Thank you for mentioning it.

src/components/OnboardingForm.tsx
iamnamananand996: If we look at it more closely, then it is required. But yes, the way it is written looks a bit wired, I should have written it in a more elegant way, without overkilling the solution
and have only `Just validate the form when the user submits it`
I was just trying to make it more user-friendly 😅. Thank you for mentioning it here.

PR #468, opened 2023-07-13
Dashboard dev

iamnamananand996:
Because

- we need metric dashboard for user to have the summary about the pipelines
- It includes some basic filter
  - Filter by days
  - filter by Pipeline status
- Dashboard also have chart representation for the triggers that happen within the time. 

This commit

- Metric dashboard


src/lib/dashboard/utils.ts
```diff
@@ -0,0 +1,316 @@
...
+        label: "Inactive",
+      };
+
+    default:
+      return {
+        variant: "default",
+        label: "Unknown",
+      };
+  }
+}
```
EiffelFly, line 209: This function has lots of unexpected side-effect. It name is getColor, but what it gets is a object with variant and label and color? This will confuse other developers

EiffelFly (commented):
@iamnamananand996 I think this PR problems centralize how you reason through your function and name. Your function is lacking a clear API and the naming is a bit inconsistent. Please go through my comment and change accordingly

src/lib/dashboard/utils.ts
```diff
@@ -0,0 +1,316 @@
...
+      amount: STATE_INACTIVE,
+      type: "pipeline",
+      change: calculatePercentageChange(
+        STATE_INACTIVE_PREVIOUS,
+        STATE_INACTIVE
+      ),
+    },
+  ];
+}
+
```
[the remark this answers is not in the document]
EiffelFly: In this case I will usually recommend using object instead of array because to use object.key is more explicit than using array[0]

iamnamananand996:
Hi @EiffelFly , I have the update the suggested changes requested above.

For `chart`, I am still thinking, shall I changes are chart completely, or make the changes current one.
I will check this [https://nivo.rocks/line/](https://nivo.rocks/line/) library, I see how much changes will be required and take the next step.

src/components/charts/LineChart.tsx
```diff
@@ -0,0 +1,141 @@
...
+                      ${pipeline_id}
+                    </div>
+                  </div>
+                  <div class="Data2" style="justify-content: flex-start; align-items: flex-start; gap: 4px; display: inline-flex">
+                    <div class="TriggerNumber" style="color: rgba(29, 36, 51, 0.80); font-size: 14px;  font-weight: 400; line-height: 20px; word-wrap: break-word">Triggers: </div>
+                    <div class="Number" style="color: #1D2433; font-size: 14px;  font-weight: 600; line-height: 20px; word-wrap: break-word">${compute_time_duration}</div>
+                  </div>
+                </div>
+              </div>
+            `;
```
EiffelFly, line 48: @iamnamananand996 Please remove unused className like Content

src/components/charts/LineChart.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  </h3>
+                </div>
+              )}
+            </>
+          )}
+        </div>
+      </div>
+    </div>
+  );
+};
```
EiffelFly, line 141: You need to remove these unused className too

src/components/skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,33 @@
...
+        className={cn(
+          "flex-1 animate-pulse bg-[#D9D9D9]",
+          animationWidth,
+          animationHeight,
+          animationClassname
+        )}
+      />
+    </div>
+  );
+};
```
EiffelFly, line 33: @iamnamananand996 We have Skeleton component in the design-system, please use it

src/components/cards/StatusCard.tsx
```diff
@@ -0,0 +1,41 @@
+import cn from "clsx";
+import React from "react";
+import { Status } from "@/types";
+import { Badge } from "../badge/Badge";
+import { Icons } from "@instill-ai/design-system";
+
+export type StatusProps = Status;
+
+const getIcon = (change: number, statusname: string) => {
```
EiffelFly, line 9: @iamnamananand996 We have discuss this before, the Status is guarded by our API clearly. So we should not use string but use the Status type provided by the backend directly!

EiffelFly (commented):
There are still changes need to be made

- statusname: string is not a good way to type thing, you should directly use the Status type provided by the backend
- Please remove all the unnecessary className

There are other things need to be changes please check the comments

EiffelFly:
@iamnamananand996 There are multiple issues that you need to fix in this PR

### The blink issue when switch the filter

The whole page will blink if you switch filter 

https://github.com/instill-ai/console/assets/57251712/f4139f88-8e86-4ec5-8aa3-5e4ef0e22a36

### The State select 

The inactive state doesn't have the card

https://github.com/instill-ai/console/assets/57251712/b8800f6d-d8a7-4b1a-a563-4da12e90953f

### The summary card didn't align with the design-doc

Especially the padding and the fontStyle

CleanShot 2023-07-16 at 22 02 38 2

EiffelFly (approved):
LGTM

PR #477, opened 2023-07-20
fix: model definition icon

iamnamananand996:
Because

- we need to add model definition icon for open AI

This commit

- add model definition icon open AI


EiffelFly (commented):
LGTM

EiffelFly (approved):

PR #492, opened 2023-07-27
feat: new table design dev

iamnamananand996:
Because

- we are adapting the new design for the table in console.

This commit

- new design for the table in console

**latest design of the table**: https://www.figma.com/file/ObcVX44RiPMdKH8RHi6n8Q/New-Console?type=design&node-id=1962-97642&mode=design&t=gXsJD6kBCMQ0ttNu-0

the PR includes the updated design of table on the below urls.

- /pipelines
- /sources
- /ais
- /blockchains
- /destinations
- /model-hub
- /dashboard
- /dashboard/pipeline/:id  (:id ---> pipeline_id)

**Note**: once the initial review is done then will start moving this to `toolkit`

**Question** regarding the filenames for `toolkit`, the current implementation is having the same name of the file which is used for the `new design` also, so while moving these changes to `toolkit`, shall I just replace everything with new implementation, or do we rename it with different name and have both version,

cc - @EiffelFly


EiffelFly (commented):
@iamnamananand996 Here are some needed changes

- Please go through all the utility class that used old design-token like instillGrey90. And change their color to corresponding color with the new design-token
- If I understand correctly, after the refactoring of the DataTable, each extended cell component like GeneralStateCell should be wrapped by the TableCell component in the design-system
- After moving into toolkit, please be careful of singular and plural of the folder name, for example `component/cell` should be `component/cells`

Here are some questions 

- There has a duplicated name "TableCell" in `cell/TableCell` and `table/TableCell`. Please clarify their usage and rename them or try to unify them (I will recommend the latter)

Here are some though that we can start from this point. 

- About utility function
 - We have lots of utility function right now, like "formatDate" in this PR. Please careful look into the toolkit to find whether there has duplicated function. 
 - Try to generate unit test for this utility function. (You can make things down very quick with the help of GPT)

EiffelFly (commented):
LGTM

EiffelFly (approved):

PR #579, opened 2023-10-03
ci(workflow): make disk free of GA hosted runner

Sarthak-instill:
Because

- We have to clean github action hosted runner so can run console image

This commit

- make disk free of GA hosted runner


EiffelFly (approved):
LGTM

PR #582, opened 2023-10-05
chore: update title onboarding

iamnamananand996:
Because

- update title onboarding

This commit

- update title onboarding


EiffelFly (approved):
LGTM

PR #643, opened 2023-10-27
ci(workflow): update docker hub login

Sarthak-instill:
Because

- we need to update the docker hub login

This commit

- update docker hub login


EiffelFly (approved):
LGTM

PR #795, opened 2023-12-14
fix: fix dashboard quota message

iamnamananand996:
Because

- fix dashboard quota message

This commit

- fix dashboard quota message


EiffelFly (approved):
LGTM

PR #814, opened 2023-12-25
chore: unused import `PageBase`

chenhunghan:
Because

- `PageBase` is not used.

This commit

- Remove the usage of `PageBase` in `404.tsx`


EiffelFly (approved):
@chenhunghan This is a good catch. Thanks!

PR #817, opened 2023-12-25
chore: remove unnecessary `error:any`, narrow to `unknown`

chenhunghan:
Because

- We have ts with strict mode on (`strict: true`), which enables [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables).`useUnknownInCatchVariables` makes `error`'s default type = `unknown`. Therefore, we do not need to allow `toastInstillError` accept `any` so it will be (mis)-used in somewhere else then the `catch` block.

This commit

- Change `error:any` to `error:unkown`, narrow the type to prevent misuse. Also remove `/* eslint-disable-next-line @typescript-eslint/no-explicit-any */`

### DIFF
--- a/packages/toolkit/src/lib/toastInstillError.ts
+++ b/packages/toolkit/src/lib/toastInstillError.ts
@@ -9,9 +9,7 @@ export function toastInstillError({
 }: {
   title: string;
   toast: UseToastReturn["toast"];
-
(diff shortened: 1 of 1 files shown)


EiffelFly (approved):
This is a very good catch, thank for your contribution!!

LGTM

PR #821, opened 2023-12-27
chore: refactor top bar navigation to make it dry, add default query options for react-query

HariBhandari07:
This commit

- Adds default options in `tanstack/react-query` config to prevent re-fetch on window focus. (I don't know if this is the required behavior for this app, but it's a good practice to have defaults and override it when required)

- Refactors Top Bar menu to use loop for menu items/links

### DIFF
--- a/apps/console/src/components/top-bar/topBarData.tsx
+++ b/apps/console/src/components/top-bar/topBarData.tsx
@@ -9,7 +9,7 @@ export const tobBarData = [
     {
         pathName: 'connectors',
         Icon: Icons.IntersectSquare,
-        name: 'IntersectSquare'
+        name: 'Connectors'
     },
     {
         pathName: 'models',
--- a/apps/console/src/components/top-bar/topBarData.tsx
+++ b/apps/console/src/components/top-bar/topBarData.tsx
@@ -1,0 +1,24 @@
+import { Icons } from "@instill-ai/design-system";
+
+export const tobBarData = [
+    {
+        pathName: 'pipelines',
+        Icon: Icons.Pipeline,
+        name: 'Pipelines'
+    },
+    {
+        pathName: 'connectors',
+        Icon: Icons.IntersectSquare,
+        name: 'IntersectSquare'
+    },
+    {
+        pathName: 'models',
+        Icon: Icons.Cube01,
+        name: 'Model Hub'
+    },
+    {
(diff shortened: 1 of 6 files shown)


EiffelFly (commented):
@HariBhandari07 Thanks for the contribution, really appreciate it

The reason we don't set the default query option is that React query already has the default retry to 3 and we think that is sufficient for our needs. Please remove the retry:2 configuration

I am also thinking about the prefetch when the window focus, which is actually very handy on some of our pages. I will look into it more, but on other pages, disabled it is a good move!

HariBhandari07:
> @HariBhandari07 Thanks for the contribution, really appreciate it
> 
> The reason we don't set the default query option is that React query already has the default retry to 3 and we think that is sufficient for our needs. Please remove the retry:2 configuration
> 
> I am also thinking about the prefetch when the window focus, which is actually very handy on some of our pages. I will look into it more, but on other pages, disabled it is a good move!

Hi @EiffelFly  I have removed the `retry` option

EiffelFly:
@HariBhandari07 I am sorry that we won't merge this PR is this release. Because it needs to be further tested. 

We will move your PR into our next release cycle (Which is after 1/2)

Thanks for your contribution, it means a lot to us

apps/console/src/components/top-bar/topBarData.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const tobBarData = [
+    {
+        pathName: 'pipelines',
+        Icon: Icons.Pipeline,
+        name: 'Pipelines'
+    },
+    {
+        pathName: 'connectors',
+        Icon: Icons.IntersectSquare,
+        name: 'IntersectSquare'
```
EiffelFly, line 12: @HariBhandari07 This will be Connectors

EiffelFly (commented):
@HariBhandari07 Hi! Sorry for the delay. We really appreciate your contribution

Would you mind fixing this additional typo for us to merge this PR for this sprint?

apps/console/src/components/top-bar/topBarData.tsx
HariBhandari07: @EiffelFly done.

EiffelFly (approved):
LGTM! @HariBhandari07 Thanks for your contribution, very appreciate it

PR #830, opened 2023-12-27
chore: add ts check for unused variables, remove unused useLayoutEffect import

HariBhandari07:
This commit

- Adds TS check for unused variables in console app
- Removes unused import of `useLayoutEffect` in `_app.tsx`


EiffelFly (approved):
@HariBhandari07 This is a really good code guard!

Happy to leverage it

PR #836, opened 2023-12-30
Fix syntax highlighting in `instill-auto-form.md`

chenhunghan:
Because

- The code block is in tsx but labeled as `ts`.

This commit

- `TypeScript` > `tsx`


EiffelFly (approved):
LGTM

PR #837, opened 2023-12-30
chore: use `mutation.isLoading` instead of tracking loading state manually

chenhunghan:
Because

- Unnecessary state `isTriggering` which overlapping `mutation.isLoading`.

This commit

- This PR replaces the manually tracked state with `mutation.isLoading` by `react-query`, see https://tanstack.com/query/v4/docs/react/guides/mutations


EiffelFly (approved):
Good catch! LGTM

PR #838, opened 2023-12-30
fix: `async` function(s) should be awaited

chenhunghan:
Because

- There are places where `Promise`(s) are not awaited. Probably not intended?

This commit

- Add missing `await`, so the message shown in the toasts is reflecting the correct state of the system.

You might want to add <https://typescript-eslint.io/rules/no-floating-promises/> to avoid this error in the future.


EiffelFly (approved):
yes this is a not-intended bug. Thanks for pointing out

I will add the eslint rule to guard this

PR #849, opened 2024-01-02
refactor: deprecate the usage of some callbacks on ReactQuery mutation

rsmelo92:
Because

According to the open issue, the maintenance of the code is reduced when react query is using callbacks instead of async functions

This commit

- Fixes it by migrating all `.mutate` to `.mutateAsync` and does the necessary adaptations
- Also, whenever I had a chance to improve readability, I did. For example, a piece of code was being repeated 5 times, so I created a functions for it, to be changed only once 
- I let just one `.mutate` as it doesnt have any callback

### DIFF
--- a/packages/design-system/tsconfig.json
+++ b/packages/design-system/tsconfig.json
@@ -1,8 +1,8 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "dist"
+    "outDir": "dist",
   },
   "include": ["**/*.ts", "**/*.tsx"],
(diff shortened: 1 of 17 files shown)


EiffelFly:
@rsmelo92 Your PR is very valuable to us, but I am sorry that we didn't review your PR in time. Could you help us rebase the main branch and we can begin the review process

rsmelo92:
for sure @EiffelFly I will do it asap!

rsmelo92:
@EiffelFly I opened a new PR to make it easier to track it

https://github.com/instill-ai/console/pull/911

PR #856, opened 2024-01-03
chore(user-profile): user profile update pipeline query

iamnamananand996:
Because

- user profile update pipeline query

This commit

- user profile update pipeline query

### DIFF
--- a/packages/toolkit/src/lib/vdp-sdk/pipeline/queries.ts
+++ b/packages/toolkit/src/lib/vdp-sdk/pipeline/queries.ts
@@ -117,7 +117,7 @@ export async function listUserPipelinesQuery(
     const pipelines: Pipeline[] = [];
 
     const queryString = getQueryString({
-      baseURL: `${userName}/pipelines?view=VIEW_FULL`,
+      baseURL: `/users/${userName}/pipelines?view=VIEW_FULL`,
       pageSize,
       nextPageToken,
       filter: null,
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/lib/vdp-sdk/pipeline/queries.ts
```diff
@@ -117,7 +117,7 @@ export async function listUserPipelinesQuery(
     const pipelines: Pipeline[] = [];
 
     const queryString = getQueryString({
-      baseURL: `${userName}/pipelines?view=VIEW_FULL`,
+      baseURL: `/users/${userName}/pipelines?view=VIEW_FULL`,
```
EiffelFly, line 120: @iamnamananand996 Because we are not sure about whether the pipeline is for user or organization and right now we all use "entityName" as this combination of identity. 

Could you leave this as it but change the names to entityName?

So the queryString will be `${entityName}/pipelines?view=VIEW_FULL`

packages/toolkit/src/view/profile/user-profile/UserProfile.tsx
```diff
@@ -36,9 +37,10 @@ export const UserProfile = () => {
     enabled: enabledQuery && entityObject.isSuccess,
   });
 
-  const pipelines = usePipelines({
+  const pipelines = useUserPipelines({
     accessToken: accessToken,
     enabled: enabledQuery,
+    userName: user.data ? user.data.id : null,
```
EiffelFly, line 43: Query should use entityName

And this entityName can get from 

```
const entityObject = useEntity()

const pipelines = useUserPipelines({
    accessToken: accessToken,
    enabled: enabledQuery && entityObject.isSuccess,
    userName: entityObject.isSuccess ? entityObject.entityName : null,
}

```

EiffelFly (commented):
@iamnamananand996 Please check the comment

EiffelFly (approved):
LGTM

PR #886, opened 2024-01-16
fix(pipeline-builder): fix the wrong icon path on the pipeline-builder's node component

iamnamananand996:
Because

- fix the wrong icon path on the pipeline-builder's node component

This commit

- fix the wrong icon path on the pipeline-builder's node component


EiffelFly:
@iamnamananand996 Sorry I didn't notice your PR and fix my working one, I will close this PR and in favor of #887

PR #889, opened 2024-01-17
fix: Typo in component configuration

jvallesm:
Because

- Component panel shows "configration" (typo)

This commit

- Fixes the typo

Found while polishing public pipelines:
CleanShot 2024-01-17 at 08 54 19


EiffelFly (approved):
LGTM! Thanks for catching this up

PR #893, opened 2024-01-22
feat: add organization support to create, clone and duplicate pipeline

iamnamananand996:
Because

- add organization support to create, clone and duplicate pipeline

This commit

- add organization support to create, clone and duplicate pipeline


packages/toolkit/src/components/ClonePipelineDialog.tsx
```diff
@@ -38,6 +39,7 @@ import { getRawPipelineRecipeFromPipelineRecipe } from "../view";
 const ClonePipelineSchema = z
   .object({
     id: z.string(),
+    accountId: z.string(),
```
EiffelFly, line 42: @iamnamananand996 We call it entity instead of account. 

I think we can call it entityId under this context

packages/toolkit/src/components/ClonePipelineDialog.tsx
```diff
@@ -86,6 +88,25 @@ export const ClonePipelineDialog = ({
...
+    userID: me.isSuccess ? me.data.id : null,
+    accessToken,
+  });
+
+  const organizationsAndUserList = React.useMemo(() => {
+    const orgsAndUserList = [];
+    if (organizations.isSuccess && organizations.data) {
+      organizations.data.map((org) => {
+        orgsAndUserList.push(org.organization);
+      });
```
EiffelFly, line 102: @iamnamananand996 Map is to use to create another list. If you are not going to use the list it creates. I will suggest using simple forEach or for...loop here

EiffelFly, line 102: @iamnamananand996 Map is to use to create another list. If you are not going to use the list it creates. I will suggest using simple forEach or for...loop here

packages/toolkit/src/view/pipeline/view-pipelines/CreatePipelineDialog.tsx
```diff
@@ -75,6 +79,28 @@ export const CreatePipelineDialog = () => {
...
+    userID: entityObject.isSuccess ? entityObject.entity : null,
+    accessToken,
+  });
+
+  const organizationsAndUserList = React.useMemo(() => {
+    const orgsAndUserList = [];
+    if (organizations.isSuccess && organizations.data) {
+      organizations.data.map((org) => {
+        orgsAndUserList.push(org.organization);
+      });
```
EiffelFly, line 93: Please use forEach instead of map if you are not going use the list it returns

EiffelFly (commented):
Function wise it's working. And the org/user lable you put in the list-item is a good ux hint. But I still want to point out there are several UX flaws that we need to fix.

## Parallel fields
These two fields, entityID(namespace) and pipeline-id are parallel field. You can find the reference here
CleanShot 2024-01-23 at 13 49 54@2x

## The full URL hint
We have a URL hint for user to better understand what is the URL looks like when they create/clone/duplicate the pipeline. We need to reflect the entityID they choose on this hint.

CleanShot 2024-01-23 at 13 57 27@2x

## When the org/user name is too long

We need to make sure even when the entityID is too long we can still handle it. We can use GitHub's way as reference (See the last one)

CleanShot 2024-01-23 at 13 59 29@2x

packages/toolkit/src/components/ClonePipelineDialog.tsx
```diff
@@ -182,43 +219,157 @@ export const ClonePipelineDialog = ({
...
+                    </div>
+                    <p className="text-semantic-fg-secondary product-body-text-3-regular">
+                      <span>Your pipeline URL in Instill AI will be:</span>
+                      <span className="ml-2 break-all product-body-text-3-semibold">
+                        {form.getValues("id") !== "" || form.getValues("id")
+                          ? `${env(
+                              "NEXT_PUBLIC_CONSOLE_BASE_URL"
+                            )}/${form.getValues(
+                              "namespaceId"
+                            )}/pipelines/${form.getValues("id")}`
```
EiffelFly, line 368: @iamnamananand996 getValues is not a reactive subscription, so it will only update the value when the whole component re-render. This is not the UX we want. 

We should use form.watch("id") to be reactive to the data

packages/toolkit/src/components/ClonePipelineDialog.tsx
```diff
@@ -182,43 +219,157 @@ export const ClonePipelineDialog = ({
...
+                                <Select.Root
+                                  value={field?.value || ""}
+                                  onValueChange={(e) => {
+                                    field.onChange(e);
+                                    if (form.getValues("id")) {
+                                      form.trigger("id");
+                                    }
+                                  }}
+                                >
+                                  <Select.Trigger className="w-full pl-[14px]">
```
EiffelFly, line 243: @iamnamananand996 We need to pick the use owned namespace (his username) as default value. In this way the user won't have weird undefined state

packages/toolkit/src/components/ClonePipelineDialog.tsx
```diff
@@ -182,43 +219,157 @@ export const ClonePipelineDialog = ({
...
+                  <div className="space-y-2">
+                    <div className="flex flex-row gap-x-4">
+                      <Form.Field
+                        control={form.control}
+                        name="namespaceId"
+                        render={({ field }) => {
+                          return (
+                            <Form.Item className="w-full">
+                              <Form.Label className="product-body-text-3-semibold">
+                                Account Name
```
EiffelFly, line 231: @iamnamananand996 We should change this to `Owner`

EiffelFly (commented):
@iamnamananand996 There are some UX issues that happened in this PR. I would recommend you test it yourself to see whether the UX is up the bar. Most of the issue can be observed before you submit the PR


https://github.com/instill-ai/console/assets/57251712/1e3ceebc-9f1f-4ba8-8ba1-135bbab7cf20

For example

- If you didn't select the owner, the URL hint fragment will be undefined, this is not good
- The URL hint is not reactive, this is not good too

EiffelFly (approved):
LGTM!

PR #911, opened 2024-01-29
refactor: deprecate the usage of some callbacks on ReactQuery mutation

rsmelo92:
Because

According to the open issue, the maintenance of the code is reduced when react query is using callbacks instead of async functions

This commit

- Fixes it by migrating all `.mutate` to `.mutateAsync` and does the necessary adaptations
- Also, whenever I had a chance to improve readability, I did. For example, a piece of code was being repeated 5 times, so I created a function for it
- I let just one `.mutate` as it doesnt have any callback

closes https://github.com/instill-ai/community/issues/455


EiffelFly (approved):
LGTM! @rsmelo92 Thanks for your contribution

PR #920, opened 2024-01-30
fix(dashboard): update pipeline chart status type

iamnamananand996:
Because

- Dashboard numbers don't add up

This commit

- updated pipeline chart `status` types


EiffelFly (approved):
LGTM

PR #930, opened 2024-01-31
chore(user-profile): update user form field with organization name and profile bio text-align

iamnamananand996:
Because

- update user form field with organization name and profile bio text-align

This commit

- update user form field with organization name and profile bio text-align


EiffelFly:
@iamnamananand996 By adapting the new breaking change this issue is fixed, I will close this PR

PR #931, opened 2024-02-01
chore(user-profile): add id to user profile bio for better URL navigation

iamnamananand996:
Because

- add id to user profile bio

This commit

- add id to user profile bio


EiffelFly:
@iamnamananand996 I think the ticket and your implementation is quite different. But I still think this add-on is valid.

PR #942, opened 2024-02-07
feat: namananand/ins 3605 add realease switch dropdown, share and delete button to view pipeline page

iamnamananand996:
Because

- add realease switch dropdown, share and delete button to view pipeline page

This commit

- add realease switch dropdown, share and delete button to view pipeline page


EiffelFly:
@iamnamananand996 There are several issues in the implementation

1. The `ReadOnlyPipelineBuilderWithVersion` component is duplicated, due to the `ReadOnlyPipelineBuilder` already expose the `recipe` and `metadata` and two of them are near identical
2. I understand this implementation is due to your want to leverage the `currentVersion` but I will suggest you not to do that due to it has some sideeffect which only make sense in the pipeline-builder edit mode
3. The version dropdown can not be closed when I click outside of it. 
4. When I delete the pipeline on the `/pipelines/pid` page, it didn't redirect me to `/pipelines` page which make the whole flow buggy

## Here is the UI deviation part

1. The version dropdown padding deviates from design
2. The default state of the dropdown's paddingX deviates from the design
3. The right eclipse button's dropdown padding deviates from the design

Please makes sure the deviations is taken care, we really want to 100% align with the design. Also make sure you test the UX interaction before submitting the PR

iamnamananand996:
Hi @EiffelFly, updated the suggested changes

EiffelFly:
@iamnamananand996 I think the implementation functionality and code quality is good. But there are some minor UX and UX issue

1. When user select the version we should close the popover
2. The blue dot should be 8px width and height according to the design handover. And it should not be compressed
CleanShot 2024-02-14 at 16 43 54
3. When the popover is selected and open the input style should be different. You can see that the border is a gray color when user open the popover
CleanShot 2024-02-14 at 16 45 07
4. The popover width should be as the same width of the popover trigger ([ref](https://www.radix-ui.com/primitives/docs/components/popover#constrain-the-content-size))

Please fix this issues accordingly

iamnamananand996:
Hi @EiffelFly , updated the changes

EiffelFly (dismissed):
LGTM

PR #945, opened 2024-02-09
chore(design-system): add storybook stories for colors

iamnamananand996:
Because

- add design-token color class to design-token

This commit

- add design-token color class to design-token


EiffelFly:
@iamnamananand996 This PR's category should be chore. And its title doesn't represent the concept well. I would suggest that we just name it as something like "Add storybook story for color......"

iamnamananand996:
Hi @EiffelFlym fixed the unit test fail

EiffelFly (approved):
LGTM

PR #947, opened 2024-02-12
chore(instill-hub): add visibility all label to dropdown

iamnamananand996:
Because

- add visibility all label to dropdown

This commit

- add visibility all label to dropdown


EiffelFly:
@iamnamananand996 We also need to set the dropdown's default value

iamnamananand996:
> @iamnamananand996 We also need to set the dropdown's default value

Hi @EiffelFly, that is already done [here](https://github.com/instill-ai/console/blob/0d294bf96244cd367f383d0f3390af51d6b424b7/packages/toolkit/src/view/pipeline/view-pipelines/ViewPipelines.tsx#L39)

EiffelFly (approved):
LGTM!

PR #955, opened 2024-02-14
fix: fix dashboard font-family issue

EiffelFly:
Because

- fix dashboard font-family issue

This commit

- fix dashboard font-family issue


EiffelFly:
@iamnamananand996 We have this font-family issue with the chart library we use. this pr fix the issue with our global font variable

PR #989, opened 2024-02-28
ci(workflows): update model ray image tag

heiruwu:
Because

- `Ray` image tag has a format change

This commit

- update `Ray` image tag variable


EiffelFly (approved):
LGTM

PR #1005, opened 2024-03-07
fix: fix alignment between input and output fields on the pipeline overview page

iamnamananand996:
Because

- namananand/ins 3640 ensure alignment between input and output fields on the

This commit

- namananand/ins 3640 ensure alignment between input and output fields on the


EiffelFly:
@iamnamananand996 

- The run button up/bottom margin is not correct

Correct one 

CleanShot 2024-03-27 at 21 32 41

Current implementation

CleanShot 2024-03-27 at 21 32 47

EiffelFly (approved):
LGTM

PR #1030, opened 2024-03-14
feat: migrate /pipelines/pid from page-route to app-route

EiffelFly:
Because

- Support Nextjs SSR
- Explore the migration path 

This commit

- migrate /pipelines/pid from page-route to app-route


apps/console/src/app/[entity]/pipelines/[id]/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  if (namespaceType === "NAMESPACE_USER") {
+    await prefetchUserPipeline({
+      pipelineName: `users/${entity}/pipelines/${id}`,
+      accessToken,
+      queryClient,
+    });
+  }
+
+  console.log(namespaceType);
```
Comment (iamnamananand996) [inline apps/console/src/app/[entity]/pipelines/[id]/page.tsx:69]: reminder to remove logs

PR #1031, opened 2024-03-19
fix: parse semi-structured/json correctly

iamnamananand996:
Because

- parse semi-structured/json correctly

This commit

- parse semi-structured/json correctly


EiffelFly:
In favor of #1033

PR #1032, opened 2024-03-19
feat: support indexable user profile page

iamnamananand996:
Because

- user indexable in search engine

This commit

- user indexable in search engine


EiffelFly (approved):
LGTM

PR #1037, opened 2024-03-26
fix(pipeline-view): set pipeline name correctly publish pipeline

iamnamananand996:
Because

- set pipeline name correctly publish pipeline

This commit

- set pipeline name correctly publish pipeline


packages/toolkit/src/view/pipeline/view-pipeline/ViewPipeline.tsx
```diff
@@ -48,6 +51,9 @@ export const ViewPipeline = () => {
     }
     setCurrentVersion(pipeline.data?.releases[0]?.id);
 
+    updatePipelineId(() => pipeline.data.id);
+    updatePipelineName(() => pipeline.data.name);
+
```
EiffelFly, line 56: @iamnamananand996 You can't do this kind of side-effect in a useMemo, that is an anti-pattern, please use another useEffect to operate this task

EiffelFly (commented):
Please check the comment

PR #1073, opened 2024-04-10
feat: add sitemap for pipelines, users and organisations

iamnamananand996:
Because

- add sitemap for pipelines, users and organisations

This commit

- add sitemap for pipelines, users and organisations


EiffelFly (approved):
LGTM

PR #1074, opened 2024-04-10
fix: update link settings

iamnamananand996:
Because

- update link settings

This commit

- update link settings


EiffelFly:
@iamnamananand996 Guard navigate has its reason to be like this way... Remove it is not a fix. I will take over this task, since its bugs originate from my implementation

PR #1114, opened 2024-04-30
fix: fix vscode settings to use tsconfigs in subfolders

orangecoloured:
Because

- vscode was throwing errors because it couldn't find the tsconfig files in nested package folders

This commit

- updates the vscode settings and eslintrc files, fixing errors


EiffelFly (approved):
LGTM!

PR #1122, opened 2024-05-03
fix: pipeline view head section clone button permission

iamnamananand996:
Because

- Login user can clone the pipeline no matter he is the owner or not

This commit

- Login user can clone the pipeline no matter he is the owner or not


EiffelFly (approved):
LGTM

PR #1127, opened 2024-05-06
fix(view-pipeline): bg fix for output field

iamnamananand996:
Because

- bg fix for output field

This commit

- bg fix for output field


EiffelFly:
@iamnamananand996 I will close this PR due to we will implement the new design

PR #1136, opened 2024-05-09
feat: update models list ui/ux

orangecoloured:
[Figma](https://www.figma.com/file/QZtPfI8NDuhEv5lPYX4z2a/Instill-Model?type=design&node-id=11%3A5322&mode=design&t=14khhkK5vHqHsh00-1)

[Eraser](https://app.eraser.io/workspace/3ZrAWVNuPMCg6QSpuD9s?origin=share)

It's a bit dirty. But most of the changes related to the models list.


orangecoloured:
@EiffelFly we can't jump to pages and their component doesn't seem to have the "1 of 2" indicator. For me it feels like replacing a couple of elements and the rest of the component will stay the same. Basically I will just replace the `Button`s. Easy. Just wondering whether it makes sense with current functionality.

EiffelFly:
> we can't jump to pages and their component doesn't seem to have the "1 of 2" indicator.

@orangecoloured Fair point, in this case we don't need to put it into design-system due to it's not universal enough. If we still want to put it into design-system then we need to think in long-term. 

- The component should stay in the composition component manner
- The component need to be future proof

That is the reason I mentioned the shacdn/ui's pagination 🦈 

I think it's on your call. We can definitely let it stay in toolkit as a small component, or we implement shacdn/ui's component in design-system and adapt it in the model list page

orangecoloured:
@EiffelFly refactored the Pagination component.

EiffelFly:
@orangecoloured Hi Alex, please make sure the PR is at least buildable 🦈

packages/toolkit/src/lib/react-query-service/model/useModels.ts
```diff
@@ -1,32 +1,57 @@
-import { useQuery } from "@tanstack/react-query";
+import { useInfiniteQuery } from "@tanstack/react-query";
 import { env } from "../../../server";
-import { listModelsQuery } from "../../vdp-sdk";
+import { Visibility, listModelsQuery } from "../../vdp-sdk";
 import type { Nullable } from "../../type";
 
 export function useModels({
```
EiffelFly, line 6: @orangecoloured Due to this behavior is different from normal useModels, we should maintain two set of queries, one is normal useModels. And another one will be `useInfiniteModels`

packages/toolkit/src/lib/vdp-sdk/model/types.ts
```diff
@@ -38,14 +40,24 @@ export type Model = {
   description: string;
   model_definition: string;
   configuration: {
-    [key: string]: any;
-  };
-  task: string;
-  state: ModelState;
+    [key: string]: string;
```
EiffelFly, line 43: @orangecoloured Do you think it's better to use Record<string, string> here?

packages/toolkit/src/view/model/CreateModelForm1.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const CreateModelSchema = z
+  .object({
+    id: z.string(),
+    description: z.string().optional().nullable(),
+    visibility: z.enum(["VISIBILITY_PRIVATE", "VISIBILITY_PUBLIC"]).default("VISIBILITY_PRIVATE"),
+    region: z.enum(["REGION_GCP_EUROPE_WEST_4"]).default("REGION_GCP_EUROPE_WEST_4"),
+    hardware: z.enum(["CPU"]).default("CPU"),
+    task: z.enum(["TASK_CLASSIFICATION", "TASK_DETECTION", "TASK_KEYPOINT", "TASK_OCR", "TASK_INSTANCE_SEGMENTATION", "TASK_SEMANTIC_SEGMENTATION", "TASK_TEXT_GENERATION", "TASK_TEXT_TO_IMAGE", "TASK_IMAGE_TO_IMAGE", "TASK_IMAGE_TO_TEXT"]).default("TASK_CLASSIFICATION"),
```
EiffelFly, line 22: @orangecoloured Could you help me extract these enum into the constant folder, in this way we can make it more maintainable

packages/toolkit/src/view/model/CreateModelForm1.tsx
```diff
@@ -0,0 +1,53 @@
...
+  /* const form = useForm<z.infer<typeof CreateModelSchema>>({
+    resolver: zodResolver(CreateModelSchema),
+    mode: "onChange",
+  }); */
+
+  console.log(modelDefinitions);
+
+  return (
+    <div>123</div>
+  )
```
EiffelFly, line 52: @orangecoloured Is this finished?

EiffelFly (commented):
@orangecoloured It seems like the feature is not ready, so I comment some code that I think it's worth discussing

packages/toolkit/src/lib/react-query-service/model/useModels.ts
orangecoloured: Should we only have `useInfiniteModels` then? It's a paginated endpoint and feels like it doesn't make sense to have it without pagination.

packages/toolkit/src/lib/vdp-sdk/model/types.ts
orangecoloured: Technically, I see no difference. But `Record` is shorter and feels more "typed". I fixed it.

packages/toolkit/src/view/model/CreateModelForm1.tsx
orangecoloured: This is finished in the other PR. This is the dirty part I mentioned. I touched some "create" stuff in this PR.

packages/toolkit/src/view/model/CreateModelForm1.tsx
orangecoloured: Let's resolve this in the next PR. Sorry I spilled some create form work into this PR.

packages/toolkit/src/lib/react-query-service/model/useModels.ts
EiffelFly: Yes we can do that before we have other use cases.

orangecoloured: Done.

packages/toolkit/src/lib/react-query-service/model/useModels.ts
```diff
@@ -1,32 +1,57 @@
-import { useQuery } from "@tanstack/react-query";
+import { useInfiniteQuery } from "@tanstack/react-query";
```
EiffelFly, line 1: @orangecoloured The filename should be renamed to `useInfiniteModels`

packages/toolkit/src/view/model/ModelHubListPageMainView.tsx
```diff
@@ -17,52 +15,125 @@ export const ModelHubListPageMainView = (
...
-        models={models.isSuccess ? models.data : []}
-        modelsWatchState={
-          modelsWatchState.isSuccess ? modelsWatchState.data : {}
+      <ModelsList
+        models={
+          models.isSuccess
+            ? models.data?.pages[pageNumber]
+              ? models.data?.pages[pageNumber].models
+              : []
+            : []
```
EiffelFly, line 126: @orangecoloured react-query will make sure once the models is success, the data is there, so you don't need to use `Optional chaining` here

packages/toolkit/src/view/model/ModelsList.tsx
```diff
@@ -0,0 +1,48 @@
...
+  onModelDelete: (
+    options?: RefetchOptions | undefined
+  ) => Promise<
+    QueryObserverResult<InfiniteData<ListUserModelsResponse, unknown>, Error>
+  >;
+  isLoading: boolean;
+};
+
+export const ModelsList = (props: ModelsListProps) => {
+  const { models, accessToken, onModelDelete, isLoading } = props;
```
EiffelFly, line 24: @orangecoloured We have accessToken and enbaedQuery in our global store. 

You can access them with useInstillStore, please look for pipeline related component for how to use it

EiffelFly (commented):
@orangecoloured I think generally speaking the code quality is nice, I left some minor stuff in the comment, please check them

packages/toolkit/src/view/model/CreateModelForm.tsx
```diff
@@ -503,9 +503,14 @@ export const CreateModelForm = (props: CreateModelFormProps) => {
           }}
           disabled={modelDefinitions ? (modelCreated ? true : false) : false}
           required={true}
-          description={`<a href="${getModelSetupGuide(
-            modelDefinition || ""
-          )}">Setup Guide</a>`}
+          description={
+            <a
+              target="_blank"
```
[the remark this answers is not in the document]
orangecoloured: This is actually completely removed in the other PR. But ok about rel. I just thought that if it's our page then it shouldn't really matter. Will be adding it in the future.

packages/toolkit/src/view/model/CreateModelForm1.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  /* const form = useForm<z.infer<typeof CreateModelSchema>>({
+    resolver: zodResolver(CreateModelSchema),
+    mode: "onChange",
+  }); */
+
+  console.log(modelDefinitions);
+
+  return <div>123</div>;
+};
```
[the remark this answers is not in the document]
orangecoloured: Could we just dismiss this file? It's removed in the other PR.
Again, sorry I left it like this.

packages/toolkit/src/view/model/ModelHubListPageMainView.tsx
orangecoloured: Do you mean to have it like this?
```
models={
  models.isSuccess
    ? models.data.pages[pageNumber]?.models
    : []
}
```

EiffelFly: yes exactly

packages/toolkit/src/view/model/ModelHubCreatePageMainView.tsx
```diff
@@ -22,6 +23,7 @@ export const ModelHubCreatePageMainView = (
         breadcrumbs={["Models", "Model Settings"]}
         className="mb-10"
       />
+      <CreateModelForm1 accessToken={accessToken} enabledQuery={enableQuery} />
```
EiffelFly, line 26: @orangecoloured I think this a dev changes, I think we need to change it back

packages/toolkit/src/view/model/ModelsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+    options?: RefetchOptions | undefined
+  ) => Promise<
+    QueryObserverResult<InfiniteData<ListUserModelsResponse, unknown>, Error>
+  >;
+  isLoading: boolean;
+};
+
+const selector = (store: InstillStore) => ({
+  accessToken: store.accessToken,
+  enabledQuery: store.enabledQuery,
```
EiffelFly, line 31: @orangecoloured If we don't need the state here, we need to remove it from the selector

packages/toolkit/src/components/card-model/CardModel.tsx
```diff
@@ -0,0 +1,111 @@
...
+} from "@tanstack/react-query";
+
+export type CardModelProps = {
+  model: Model;
+  accessToken: Nullable<string>;
+  onDelete: (
+    options?: RefetchOptions | undefined
+  ) => Promise<
+    QueryObserverResult<InfiniteData<ListUserModelsResponse, unknown>, Error>
+  >;
```
EiffelFly, line 36: @orangecoloured I think this callback return is a bit confusing, I noticed that you try to use `query.refetch`

We can make it more general, just like (options) => void

and you use it as `() => query.refetch`

But if you want this prop to be only used for refetch then we should rename it to be more specific. I am ok with either way 🦤

packages/toolkit/src/view/model/CreateModelForm1.tsx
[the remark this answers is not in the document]
EiffelFly: But since this PR is merged into the main, and since you already have this one another PR, I think we can safely delete it?

EiffelFly (commented):
@orangecoloured I think the code wise it's good now, I left some comment and I will begin to review the functionality

packages/toolkit/src/view/model/CreateModelForm1.tsx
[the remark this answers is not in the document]
orangecoloured: Alright. Makes sense. Done.

packages/toolkit/src/view/model/ModelHubCreatePageMainView.tsx
orangecoloured: I removed it.

packages/toolkit/src/components/card-model/CardModel.tsx
orangecoloured: I simplified it to `() => void`. If we need to pass any options we can always update the type.

EiffelFly (commented):
@orangecoloured I left some comments please check them. Once Heiru merged his PR into main I will begin to review the functionality and style

packages/toolkit/src/components/card-model/CardModel.tsx
```diff
@@ -0,0 +1,95 @@
...
+export type CardModelProps = {
+  model: Model;
+  accessToken: Nullable<string>;
+  onDelete: () => void;
+};
+
+export const CardModel = (props: CardModelProps) => {
+  const { model, accessToken, onDelete } = props;
+  const params = useParams();
+  const entity = params.entity ? String(params.entity) : null;
```
EiffelFly, line 32: @orangecoloured We don't directly interpret entity like this since there will be entity for user and organization. That is the reason we rely on useAppEntity to handle this stuff 🦤

EiffelFly (commented):
@orangecoloured Final comment!! We are very close 🦤

EiffelFly:
@orangecoloured There is one issue that need to be solved in this PR, it's about the visibility of model on the /entity/models page.

## Let's make it act like this way 

- When the viewer is exact the owner of the current entity, the /entity/models page should display all the model owned by the entity
- When the viewer is not the owner of the current entity, the /entity/models page should display the public models owned by the entity

## For example

- Current login user's entity is summerbud, I belong to the org instill-ai
- When I visit /summerbud/models -> Display all the models I own
- When I visit /instill-ai/models -> Display all the models instill-ai own
- When I visit /alex/models -> Display all the models Alex own
- When I visit /alex-corp/models -> Display all the public models Alex own

## How to implement this

We have a similar endpoint for pipelines https://github.com/instill-ai/console/blob/121c5704f036db205808e6bd1a6e1de8b93f1530/packages/toolkit/src/lib/vdp-sdk/pipeline/queries.ts#L114

So you will need additional endpoint which query the models under specific entity like `getUserModelsQuery`

At the same time we need to make it act as infinite-query to enable current UX on page

EiffelFly:
@orangecoloured Some UI issues I discovered

## We need to make the search and visibility selection fixed width
according to the design handover they are both `w-[300px]`

https://github.com/instill-ai/console/assets/57251712/b415d44b-a380-47c6-9010-f3d3a4083206

## The create model 
- The create model page's content should has a fixed width 
- The margin of this page is not correct. 
- The `Create a model` font style is not correct
- When I expand the width of the screen the content should has fixed width and centered

https://github.com/instill-ai/console/assets/57251712/0613dd40-e434-40b4-a35a-f6df5ee7be9a

## The version tab placeholder
There has some flicker happening when the version is loading


https://github.com/instill-ai/console/assets/57251712/b5d44ba6-145e-4e05-a43d-a320f523cc63

EiffelFly:
@orangecoloured I forgot to mention the URL structure might need to align with the tab structure

For example, when the user switch to version tab the URL should be /summerbud/models/mid/versions

So user can quicker navigate to it from different space

EiffelFly (commented):
Hi Alex, I only have this one question left

packages/toolkit/src/components/ImageWithFallback.tsx
```diff
@@ -21,6 +21,16 @@ export const ImageWithFallback = ({
...
+  const currentImageSrc = React.useRef<string>(src);
+
+  React.useEffect(() => {
+    if (currentImageSrc.current !== src) {
+      currentImageSrc.current = src;
+
+      setError(false);
+    }
+  }, [src]);
+
```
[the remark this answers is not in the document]
orangecoloured: The use case:

We have an image which is broken and the component switches into error mode. We update the image, for example using a modal or some other way wich does not require to change routes or dismount the component. Being in the error mode nothing will change and to see the updated image we need to reset the component, either refresh the page or switch views etc. With this logic as soon as the image component receives a new image src, it tries to render the image again. And if succeedes - we see an updated image, if not - it falls back into the error mode.

EiffelFly (dismissed):
LGTM

EiffelFly (approved):
LGTM

PR #1138, opened 2024-05-10
feat: update model creation page

orangecoloured:
[Figma](https://www.figma.com/file/QZtPfI8NDuhEv5lPYX4z2a/Instill-Model?type=design&node-id=11%3A5322&mode=design&t=14khhkK5vHqHsh00-1)
[Eraser](https://app.eraser.io/workspace/3ZrAWVNuPMCg6QSpuD9s?origin=share)

Because

- implements new api and designs for the model creation view

This commit

- new model creation form + request


package.json
```diff
@@ -39,5 +39,5 @@
   "engines": {
     "node": ">=20.0.0"
   },
-  "packageManager": "pnpm@8.8.0"
-}
\ No newline at end of file
+  "packageManager": "pnpm@8.8.0+sha512.798e6b322669ce63c8da856bd62ad1f7b6f36f4dfaa0ac025af2bdd700d09dd5dca943e5cad3edac5d1b3baebc4b2c3fb80fbb8537f936733c32be2ec7805144"
+}
```
EiffelFly, line 43: @orangecoloured I think we don't need to specify sha here

packages/toolkit/src/lib/vdp-sdk/model/mutations.ts
```diff
@@ -63,90 +69,31 @@ export type CreateUserArtiVCModelPayload = {
...
+    const client = createInstillAxiosClient(accessToken, true);
+
+    const { data } = await client.post<CreateUserModelResponse>(
+      `/${isOrg ? "organizations" : "users"}/${entityName}/models`,
+      payload
+    );
+
+    return Promise.resolve(data.model);
+  } catch (err) {
+    return Promise.reject(err);
```
EiffelFly, line 96: @orangecoloured I think you might confused around entityName and entity.id

Let's take my account for example, its entity name will be `users/summerbud` and it entity id will be `summerbud`

If it's a organization then its entity name might be `organizations/instill-ai` and its entity id is `instill-ai`

I think we can follow how we write the sdk in other services, instead of exposing flag like isOrg, we only need to expose entityName

EiffelFly (commented):
@orangecoloured Please help me move the create model modal into this PR and I left some comment please check them

package.json
orangecoloured: It was some kind of a mess up. Fixed now.

packages/toolkit/src/lib/vdp-sdk/model/mutations.ts
orangecoloured: Done

PR #1145, opened 2024-05-13
feat: update input and output fields design

iamnamananand996:
Because

- update input and output fields on design

This commit

- update input and output fields on design


iamnamananand996:
Hi @EiffelFly, this PR is ready for review

PR #1149, opened 2024-05-14
fix: secret deletion glitch

thewbuk:
Because

- When typing during deleting secret, the modal UI disappears from time to time. - **Could not replicate** 
- Change "Please type users/user-id/secrets/secret-id" to "Please type secret-id to confirm."

This commit

- Change "Please type users/user-id/secrets/secret-id" to "Please type secret-id to confirm."


EiffelFly:
@thewbuk the prettier format is failed, you can run `pnpm format` in your root folder to fix it. @orangecoloured helped us add this command in the lint-staged in one of his pr. Once we merged it, it will be failed-free. But before that you still need to manually do it

EiffelFly:
@thewbuk There is a bug on e2e test caused your PR's e2e test is failed, please help me rebase main and push to remote

PR #1151, opened 2024-05-14
feat: create hub page

thewbuk:
[Figma](https://www.figma.com/design/eTh2TCwmqN4MxiHJ4vTnPH/Instill-Hub?node-id=1526-36267&m=dev)

Because

- Implements new hub page

This commit

- New hub page, new "What's new" and "Newsletter" components, Featured pipelines, username display, pipelines count etc

TODO BIG:

- [x] **Connect to backend for "featured" pipelines**. _Waiting for backend to be populated_ 

TODO SMALL: 

- [ ] Sorting works but selecting some combinations does not show both ticks 
- [x] Blogs are not sorted correctly and the date has the wrong format
- [x] Change next/previous icons in blog component 
- [ ] Implement remark formatter 
- [ ] Use current `formatDate.ts` (if possible) 
- [x] Update icons in dropdown 
- [ ] Change bg below divider 
- [ ] Divider full width of page
- [x] Add link to blog post on image and description. **Important:** Currently using slug attributes from [mdx files](https://github.com/instill-ai/instill.tech/tree/main/blog). If the slug does not correspond to URL the link will fail.


packages/toolkit/src/components/LatestChangesCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+          .sort(
+            (a, b) =>
+              new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+          );
+        setChangelogs(filteredAndSortedChangelogs.slice(0, 3));
+      })
+      .catch((error) => {
+        console.error("Error fetching changelogs:", error);
+      });
+  }, []);
```
EiffelFly, line 36: @thewbuk This will make the console request this URL every re-render which is not good.

Please re-write this quest under a react-query useQuery function and utilize the quest hook to get the data, this case the data will be cached

EiffelFly (commented):
@thewbuk The request on productland and the request to GitHub both have the problem I mentioned in the comment, please re-write it into the react-query hook to leverage the cache system

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,172 @@
...
+          <Icons.ArrowRight
+            className={`h-6 w-6 ${currentIndex === blogPosts.length - 1 ? "stroke-gray-400" : "stroke-[#1D2433CC]"}`}
+          />
+        </button>
+      </div>
+    </div>
+  );
+};
+
+export default NewsLetterCard;
```
EiffelFly, line 172: Hi @thewbuk Let talk about the design token. When you browse our react component you can find out that we didn't use any original color scheme from Tailwind. And use our owned design-token like text-semantic-fg-primary. 

You can actually find them in the handover on Figma. Some time it will be the semantic value like this, but sometime when Dani runs out of time, he may give the color as netural-100, you can reverse check it from our design-token file 

https://github.com/instill-ai/console/blob/main/packages/design-tokens/tokens/semantic/colour.json

EiffelFly (commented):
@thewbuk Please double check and replace all the Tailwind color token with our design-token 🦤

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,202 @@
...
+    return "";
+  }
+};
+
+const NewsLetterCard = () => {
+  const { data: blogPosts = [], isLoading, isError } = useQuery({
+    queryKey: ["blogPosts"],
+    queryFn: fetchBlogPosts,
+    retry: 3,
+  });
```
EiffelFly, line 100: @thewbuk Because we don't know whether we will re-use this or not, with the perspective of maintainability and as a guidance for future development, could you help us centralize this hook into react-query-service folder, you can create another folder like `misc` and put it there

EiffelFly (commented):
@thewbuk I have some further comment on react-query

packages/toolkit/src/components/LatestChangesCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+              })
+              : new Date(changelog.updatedAt).toLocaleDateString("en-US", {
+                year: "numeric",
+                month: "long",
+                day: "numeric",
+              })}
+          </button>
+          <p>{changelog.title}</p>
+        </div>
+      ))}
```
EiffelFly, line 48: @thewbuk When utilize reqct-query. The correct way to make sure data is there is like

const changelogs = useChangelogs()

and you only need to check whether the query is success something like

```
if (changelogs.isSuccess){
  return changelogs.data.map() // <---- Don't need to do the optional chain anymore
}

```

packages/toolkit/src/components/LatestChangesCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+        rel="noopener noreferrer"
+      >
+        <button
+          type="button"
+          className="my-auto ml-2 mt-4 flex justify-start text-semantic-accent-default product-button-button-2 hover:!underline"
+        >
+          View changelog
+          <Icons.ChevronRight className="my-auto h-4 w-4 stroke-semantic-accent-default" />
+        </button>
+      </a>
```
EiffelFly, line 61: @thewbuk anchor element should not have interactive element inside, or this will violate W3C HTML protocol. You either remove the button and use div there, or use button directly and router.push

> The [a](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) element can be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g., buttons or other links). This example shows how this can be used to make an entire advertising block into a link:

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+  if (isError) {
+    return <div>Error fetching blog posts.</div>;
+  }
+
+  const { imageUrl, title, publishedOn, themeImgAlt, slug } =
+    blogPosts[currentIndex];
+
+  return (
+    <div className="flex h-[450px] flex-col gap-y-2 rounded-sm border border-semantic-bg-line p-4">
+      <h2 className="mb-4 text-2xl font-bold">What's New?</h2>
```
EiffelFly, line 56: @thewbuk Hi I forget to mention that we have design-token for typography too. 

Take this section for example, in design handover, the value is set to `Prod/Headings/Heading 3` the token will be `product-headings-heading-3`.

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+            className="h-full w-full object-cover"
+          />
+        </a>
+      </div>
+      <button
+        type="button"
+        className="my-1 w-min whitespace-nowrap rounded-sm bg-semantic-accent-bg px-2 py-2 capitalize text-semantic-accent-default product-button-button-2 hover:!underline "
+      >
+        {publishedOn}
+      </button>
```
EiffelFly, line 80: @thewbuk Although the design-handover have this as button, but since it didn't has any button functionality, we can replace the button element with div and utilize the button style like this 

```
<div className={cn(buttonVariants({ variant: "secondaryColour", size: "sm" }))}> 
```

A side note, when it comes to button, in the desgin-handover it will reference its variant, and we have a unified button in the design-system, please use that component first 🦤

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+              }`}
+          />
+        </button>
+        <button
+          type="button"
+          onClick={handleNext}
+          className={`${currentIndex === blogPosts.length - 1
+            ? "cursor-not-allowed text-gray-400"
+            : ""
+            }`}
```
EiffelFly, line 107: @thewbuk You can use `text-semantic-fg-disabled` here to address the disabled status

packages/toolkit/src/components/card-pipeline-hub/Body.tsx
```diff
@@ -0,0 +1,81 @@
...
+          type="button"
+          className="my-auto !normal-case text-semantic-accent-default product-button-button-2 hover:!underline"
+          onClick={() => {
+            router.push(
+              `/${pipeline.owner_name.split("/")[1]}/pipelines/${pipeline.id}`
+            );
+          }}
+        >
+          {pipeline.id}
+        </button>
```
EiffelFly, line 57: There are lots of button that is not used the button in the design-system, please help us replace them

packages/toolkit/src/components/card-pipeline-hub/CardPipeline.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Body.Skeleton />
+    </div>
+  );
+};
+
+export const CardPipeline = (props: {
+  ownerID: string;
+  pipeline: Pipeline;
+  isOwner: boolean;
+  disabledPermissionLabel?: boolean;
```
EiffelFly, line 20: @thewbuk This seems unused variable

packages/toolkit/src/components/card-pipeline-hub/Head.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const { toast } = useToast();
+  const { accessToken } = useInstillStore(useShallow(selector));
+
+  const deletePipeline = useDeleteUserPipeline();
+  async function handleDeletePipeline() {
+    try {
+      await deletePipeline.mutateAsync({
+        pipelineName: pipeline.name,
+        accessToken: accessToken ? accessToken : null,
+      });
```
EiffelFly, line 51: When handle the mutation, we should block the whole process when the accessToken is null

The reason we mark the accessToken as nullable string is to have flexibility on working with some edge case 🦅

packages/toolkit/src/components/card-pipeline-hub/Head.tsx
```diff
@@ -0,0 +1,136 @@
...
+    return null;
+  }, [pipeline]);
+
+  const displayName = React.useMemo(() => {
+    const owner = pipeline.owner as UserOwner;
+    const username = owner?.user?.name?.split("/")[1]; // Get the part after "users/"
+    const ownerName =
+      owner?.user?.profile?.company_name ||
+      owner?.user?.profile?.display_name ||
+      "";
```
EiffelFly, line 88: @thewbuk In our product, due to we follow the Google AIP resource oriented api design. All the `name` have very unique meaning, which means the full path toward the resource like `users/uid`. So I will recommend change the naming here

- `username` -> `userID`
- `ownername` -> `ownerDisplayName`

Please don't be afraid to have long variable name in this repo, if we can make the meaning clear, I think that is a good practice

packages/toolkit/src/components/card-pipeline-hub/Head.tsx
```diff
@@ -0,0 +1,136 @@
...
+            className="my-auto !normal-case text-semantic-accent-default product-button-button-2 hover:!underline"
+            onClick={() => {
+              router.push(`/${ownerID}`);
+            }}
+          >
+            {ownerID}
+          </button>
+          <div className="rounded-m bg-neutral-50 rounded-full border border-neutral-200 p-1 text-center text-sm font-semibold text-black">
+            {displayName}
+          </div>
```
EiffelFly, line 129: @thewbuk You can use the Tag component in the design-system here

packages/toolkit/src/components/card-pipeline/Body.tsx
```diff
@@ -32,19 +49,43 @@ export const Body = ({ pipeline }: { pipeline: Pipeline }) => {
         >
           {pipeline.id}
         </button>
+        {pipeline.isFeatured && (
+          <button
+            type="button"
+            className="my-auto rounded-md bg-[#F8F5FF] p-2 !normal-case text-[#6E35DE] product-button-button-2 hover:!underline"
+          >
+            Featured
+          </button>
```
EiffelFly, line 58: @thewbuk You can use Tag here

packages/toolkit/src/components/card-pipeline/Body.tsx
```diff
@@ -32,19 +49,43 @@ export const Body = ({ pipeline }: { pipeline: Pipeline }) => {
...
-        pipelineName={pipeline.name}
-        recipe={pipeline.recipe}
-        metadata={pipeline.metadata}
-        className="h-[250px] w-full !border-none !px-0"
-      />
+      <div className="flex w-full justify-end px-3 pb-3">
+        <p className="line-clamp-3 text-xs font-normal text-gray-400">
+          TODO Clone
+        </p>
+      </div>
```
EiffelFly, line 86: @thewbuk Will this be ready in this release?

packages/toolkit/src/components/top-bar/CETopbarDropdown.tsx
EiffelFly, line 1: @thewbuk I think we didn't explain this section to you will, this is a big feature that will help user switch their current working entity. I will handle this part. We can rollback this change. 

Side note: If you encounter some design element that you can not grasp the full functionality, I am here to help you clarify it 🐼

EiffelFly (commented):
@thewbuk Please check the comments I left 👍

packages/toolkit/src/components/NewsLetterCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+            fallbackImg={
+              <Icons.Box className="h-8 w-8 stroke-semantic-fg-primary" />
+            }
+            className="h-full w-full object-cover"
+          />
+        </a>
+      </div>
+      <div className={`my-1 w-min whitespace-nowrap rounded-sm  px-2 py-2 capitalize hover:!underline ${cn(
+        buttonVariants({ variant: "secondaryColour", size: "md" })
+      )}`}>
```
EiffelFly, line 79: @thewbuk This should be rewritten as 

```suggestion
      <div className={cn(buttonVariants({ variant: "secondaryColour", size: "md" }), "your additional className")}>
```

packages/toolkit/src/components/card-pipeline-hub/Body.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useRouter } from "next/navigation";
+
+const BodySkeleton = () => {
+  return (
+    <div className="flex w-full flex-col px-2">
+      <div className="mb-2 h-5 w-1/2 animate-pulse rounded bg-semantic-bg-secondary" />
+      <div className="h-[250px] w-full animate-pulse bg-semantic-bg-secondary" />
+    </div>
+  );
+};
```
EiffelFly, line 14: @thewbuk We have an unified Skeleton component, you can try to use it

packages/toolkit/src/components/card-pipeline-hub/Body.tsx
```diff
@@ -0,0 +1,80 @@
...
+          }}
+        >
+          {pipeline.id}
+        </Button>
+        {pipeline.isFeatured && (
+          <Tag
+            className="my-auto rounded-md bg-semantic-secondary-bg  p-2 !normal-case text-semantic-secondary-on-bg product-button-button-2 hover:!underline"
+          >
+            Featured
+          </Tag>
```
EiffelFly, line 63: @thewbuk Tag has its own variant, please take a look at how we use it in other file

EiffelFly (commented):
@thewbuk I have left some comment and please help me rebase main first, there are some file that already merged but show up in this PR

EiffelFly:
@thewbuk When I begin to validate the style, I immediately discover that there have many stuffs deviate from the design handover. 

1. The discord/feature banner's position is wrong
2. The what's new button is wrong

Please make sure your result is completely align with the design-handover, if you find any issue you can always bring that up and discuss a workaround. 

### Design handover
VDP Hub - v1 - Featured

### Current implementation
CleanShot 2024-05-17 at 14 12 24

PR #1158, opened 2024-05-16
feat: new model page

orangecoloured:
Because

- implementing new model page UI/UX

This commit

- top section
- versions tab
- settings tab


packages/design-system/src/ui-helpers/getModelInstanceTaskToolkit.tsx
```diff
@@ -178,12 +166,7 @@ export const getModelInstanceTaskToolkit = (task: string) => {
...
-            <Icons.TextGenerationChat
-              className={cn(
-                "my-auto h-5 w-5 stroke-semantic-fg-primary",
-                className
-              )}
-            />
+            <Icons.TextGenerationChat className={className ?? "h-4 w-4"} />
           );
         },
         label: "Text Generation Chat",
```
EiffelFly, line 172: @orangecoloured This way the user of this component may need to think he might completely overwrite something. That is the reason why we usually have a default value and then user can partially overwrite it. This way we can reduce the mental overhead of user

packages/toolkit/src/components/StateLabel.tsx
EiffelFly, line 1: @orangecoloured I forget to mention, something like `instillGrey50` is the old token we use, could you help me double check the new token and replace them?

EiffelFly (commented):
@orangecoloured I will sequentially add new comment~

packages/design-system/src/new-ui/DataTable/DataTable.tsx
```diff
@@ -28,8 +29,12 @@ interface DataTableProps<TData, TValue> {
...
   loadingRows: Nullable<number>;
   children?: React.ReactNode;
-  primaryText: Nullable<string>;
-  secondaryText: Nullable<string>;
+  primaryText?: Nullable<string>;
+  secondaryText?: Nullable<string>;
+  manualPagination?: boolean;
+  pageCount?: number;
+  onPaginationChange?: (state: PaginationState) => void;
+  pagination?: PaginationState;
```
EiffelFly, line 37: @orangecoloured I will recommend we use paginationState and onPaginationStateChange.

In this codebase, we believe the exchange between longer context with better readability is beneficial. So don't be afraid to have long variable name. Most of time it is very helpful for maintainer

packages/design-system/src/new-ui/DataTable/DataTable.tsx
EiffelFly, line 1: @orangecoloured There are other resuable component in shacdn/ui. After we release this model pages. Could you take a look at how we can better make use of them?

https://ui.shadcn.com/docs/components/data-table#reusable-components

EiffelFly: I still don't think DataTable component is good to use even with shacdn/ui's template, it lacks of composition component flexibility. For example, what if I want to change the style of the pagination button. It is fixed style right now. Please also put this into consideration in the next step

packages/toolkit/src/components/StateIcon.tsx
EiffelFly, line 1: Could you help us replace the color with our own design-token 🦤

packages/toolkit/src/components/StateLabel.tsx
EiffelFly, line 1: These color tokens need to be changed too

packages/toolkit/src/lib/react-query-service/model/onSuccessAfterModelMutation.ts
```diff
@@ -93,7 +93,7 @@ export async function onSuccessAfterModelMutation(
             [model.name]: watch,
           }
         : { [model.name]: watch }
-    );
+    ); */
```
EiffelFly, line 96: @orangecoloured We can safely remove these lines since they are old and deprecated code 🦤

packages/toolkit/src/lib/react-query-service/model/useInfiniteModelVersions.ts
```diff
@@ -0,0 +1,70 @@
...
+export function useInfiniteModelVersions({
+  accessToken,
+  enabledQuery,
+  modelName,
+  pageSize,
+  retry,
+}: {
+  accessToken: Nullable<string>;
+  enabledQuery: boolean;
+  modelName: string;
```
EiffelFly, line 21: @orangecoloured Most of time, to ensure we have more flexibility on the outer space, I will put this modelName as Nullable<string> but guard it below like

```
if (modelName && enabledQuery) {
    enabled = true;
  }
```

packages/toolkit/src/lib/vdp-sdk/model/mutations.ts
```diff
@@ -36,6 +36,16 @@ export type CreateUserModelPayload = {
   configuration: Record<string, string>;
 };
 
+export type UpdateUserModelPayload = {
+  description?: string;
+  source_url?: string;
+  documentation_url?: string;
+  license?: string;
+  //visibility: Visibility;
```
EiffelFly, line 44: @orangecoloured What happened with the visibility?

packages/toolkit/src/view/model/ModelHubSettingPageMainView.tsx
```diff
@@ -152,7 +198,7 @@ export const ModelHubSettingPageMainView = (
       <ModelConfigurationFields
         model={model.isSuccess ? model.data : null}
         marginBottom="mb-10"
```
EiffelFly, line 200: @orangecoloured If you think we will never use this code, we can safely remove them

packages/toolkit/src/view/model/view-settings/ModelSettingsContentViewer.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const ModelSettingsContentViewer = ({
+  selectedTab,
+  model,
+  onUpdate,
+}: ContentViewerProps) => {
+  if (!model) {
+    return null;
+  }
```
EiffelFly, line 18: @orangecoloured If the model is not there, is it due to the query is still loading?

packages/toolkit/src/view/model/view-settings/ModelSettingsHead.tsx
```diff
@@ -0,0 +1,171 @@
...
+export type HeadProps = {
+  selectedTab: ModelSettingsTabs;
+  onTabChange: React.Dispatch<React.SetStateAction<ModelSettingsTabs>>;
+  model?: Model;
+};
+
+const OWNER = {
+  avatarUrl: "",
+  name: "",
+};
```
EiffelFly, line 30: @orangecoloured To avoid confusing, I am very hesitant to use empty string in this codebase, most of the time I will use null to represent empty value. I will recommend this act due to it can shape a pre "Failed Early". So whenever there has bug we can trace it quicker

packages/toolkit/src/view/model/view-settings/ModelSettingsVersions.tsx
```diff
@@ -0,0 +1,151 @@
...
+  ];
+
+  const versions = useInfiniteModelVersions({
+    accessToken,
+    enabledQuery,
+    modelName: model.name,
+  });
+
+  useEffect(() => {
+    if (!versions.data?.pages[pagination.pageIndex]) {
```
EiffelFly, line 99: @orangecoloured Please use isSuccess to validate whether the data is there 🦤

EiffelFly (commented):
@orangecoloured Please check the comment I left

packages/design-system/src/new-ui/DataTable/DataTable.tsx
orangecoloured: Should we just update the `DataTable` component along the way with the features needed?
Like I did in this PR.

EiffelFly (commented):
@orangecoloured After you fix all the other issue, let merge this into the big model PR

packages/toolkit/src/lib/vdp-sdk/model/mutations.ts
orangecoloured: It can't be changed yet via `PATCH`. @heiruwu said that it's gonna change soon and we'll put it back there.

packages/toolkit/src/view/model/view-settings/ModelSettingsContentViewer.tsx
orangecoloured: I updated the code to show spinner while the model is loading

packages/toolkit/src/lib/vdp-sdk/model/mutations.ts
EiffelFly: got it!! Thanks for clarification

EiffelFly (commented):
@orangecoloured Let speed thing up, once you fix this one we will merge it into big Model PR

PR #1159, opened 2024-05-16
feat: 3939 dynamically generated og image for pipelines and user page

iamnamananand996:
Because

- dynamically generated og image for pipelines and user page

This commit

- dynamically generated og image for pipelines and user page


iamnamananand996:
Hi @EiffelFly, this PR is ready for review

PR #1168, opened 2024-05-20
fix(delete_secret_modal): secret deletion glitch

thewbuk:
Because

- When typing during deleting secret, the modal UI disappears from time to time. - **Could not replicate** 
- Change "Please type users/user-id/secrets/secret-id" to "Please type secret-id to confirm."

This commit

- Change "Please type users/user-id/secrets/secret-id" to "Please type secret-id to confirm."


EiffelFly (approved):
LGTM

PR #1171, opened 2024-05-21
fix: change name case in pipeline to normal case

thewbuk:
Because

- Change Display Name in pipeline to normal case

### DIFF
--- a/packages/toolkit/src/view/hub/hub-view/card-pipeline-hub/Head.tsx
+++ b/packages/toolkit/src/view/hub/hub-view/card-pipeline-hub/Head.tsx
@@ -39,24 +39,37 @@ export const Head = ({
     return null;
   }, [pipeline]);
 
+    // Function to transform the ID into a more readable format
+    const transformID = (id: string) => {
+      return id
+        .split(/[-_]/) // Split the ID by dash and underscore
+        .map((part) => part.charAt(0).toUpperCase() + part.slice(1)) // Capitalize each part
+        .join(" "); // Join the parts with a space
+    };
+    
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/view/hub/hub-view/card-pipeline-hub/Head.tsx
```diff
@@ -39,24 +39,37 @@ export const Head = ({
...
+      .join(" "); // Join the parts with a space
+  };
+
   const displayName = React.useMemo(() => {
     const owner = pipeline.owner;
 
     if ("user" in owner) {
-      // Owner is of type user
       const userOwner = owner as UserOwner;
       const userID = userOwner.user.name.split("/")[1]; // Get the part after "/"
```
EiffelFly, line 55: @thewbuk you can also do something like this
```suggestion
      const userID = userOwner.user.id
```

packages/toolkit/src/view/hub/hub-view/card-pipeline-hub/Head.tsx
```diff
@@ -39,24 +39,37 @@ export const Head = ({
     return null;
   }, [pipeline]);
 
+  // Function to transform the ID into a more readable format
+  const transformID = (id: string) => {
```
EiffelFly, line 43: @thewbuk I will suggest use a more verbal name, something like transformIDToCapitalizeName

EiffelFly (commented):
@thewbuk Please check these comments 🦤

EiffelFly (approved):
LGTM

PR #1174, opened 2024-05-23
fix: adjust search and visibility options style in Pipelines 

thewbuk:
Because

- Adjusts style of Search Pipelines and Visibility in Pipelines page to match the Hub page


EiffelFly (approved):
LGTM

PR #1179, opened 2024-05-24
feat: remove name sorting option and fix asc/desc in sorting 

thewbuk:
Because
- Remove name sort option
- Choosing Ascending sends Descending to the backend and vice versa. This is a workaround to be discussed


EiffelFly:
@thewbuk Please take a look the https://github.com/instill-ai/console/pull/1179/commits/c3b59cb264dba11f3ab00ecc02a70913c71f5f44

The key takeaways are

- The react-query has its own internal update cycle, in your previous implementation, the refetch and the update state pattern are in the same cycle, thus your refetch is still using the previous cycle's state. So the error happened
- To correctly control the react-query cache, you need to use the queryKey to control it.

PR #1187, opened 2024-05-30
feat: add total runs sort option

thewbuk:
Because

- In Hub add sorting by Total Runs


EiffelFly (commented):
@thewbuk The logic is quite the opposite, please make sure the query is correct

https://github.com/instill-ai/console/assets/57251712/27bebb0b-8b8a-4854-9180-d9d6aec7d345

PR #1203, opened 2024-06-05
docs(license): update repo license

pinglin:
Because

- Instill Core project is now with Instill Community License

This commit

- update license file


EiffelFly (approved):
LGTM

PR #1212, opened 2024-06-07
fix(model-forms): fix model forms input descriptions font size

orangecoloured:
Because

- the model create and edit forms had wrong input description font size

This commit

- fixes create and edit forms input description font size


EiffelFly (approved):
LGTM

PR #1213, opened 2024-06-07
fix(newsletter): use latest tag to fetch blogs

thewbuk:
Because

- Use latest tag to fetch newsletters discarding drafts


packages/toolkit/src/lib/react-query-service/misc/useBlogPosts.ts
```diff
@@ -46,8 +44,14 @@ const formatDate = (dateString: string | undefined) => {
 };
 
 const fetchBlogPosts = async () => {
+  // Fetch the latest release tag
+  const tagsResponse = await axios.get(
+    "https://api.github.com/repos/instill-ai/instill.tech/tags"
+  );
+  const latestTag = tagsResponse.data[0].name;
```
orangecoloured, line 51: Should it be `tagsResponse.data[0]?.name`? In case there are no posts.

packages/toolkit/src/lib/react-query-service/misc/useBlogPosts.ts
```diff
@@ -69,6 +73,9 @@ const fetchBlogPosts = async () => {
           metadata[key] = value;
         }
       });
+
+      const draft = metadata.draft === "true" || false;
```
orangecoloured, line 77: I think it can be simplified to `const draft = metadata.draft === "true"`. It's gonna evaluate to either `true` or `false`.

packages/toolkit/src/lib/react-query-service/misc/useBlogPosts.ts
```diff
@@ -79,6 +86,7 @@ const fetchBlogPosts = async () => {
         publishedOn: formatDate(metadata.publishedOn?.replace(/^"|"$/g, "")),
         themeImgAlt: metadata.themeImgAlt || "Blog post image",
         slug: file.path.replace(/^blog\//g, "").replace(/\.mdx$/g, ""),
+        draft: draft,
```
orangecoloured, line 89: This can be just `draft`. Without the explicit assignment `draft: draft`.

orangecoloured (changes requested):

packages/toolkit/src/lib/react-query-service/misc/useBlogPosts.ts
thewbuk: If that's the case should I then check for the next tag?

orangecoloured (approved):

packages/toolkit/src/lib/react-query-service/misc/useBlogPosts.ts
orangecoloured: yeah, ideally `${latestTag ? `?ref=${latestTag}` : ''}`

EiffelFly (approved):
LGTM

PR #1219, opened 2024-06-08
feat(hub): add direct routes for explore/featured + pipelines/models

orangecoloured:
Because

- we needed a way to directly share the url for explore/featured models/pipelines pages

This commit

- now the direct urls are shareable + we remember which page was visited last, so coming to /hub will seamlessly redirect to last visited section

### DIFF
--- a/apps/console/src/app/hub/[type]/[list]/page.tsx
+++ b/apps/console/src/app/hub/[type]/[list]/page.tsx
@@ -1,6 +1,5 @@
 import { Metadata } from "next";
 import { HubPageRender } from "./render";
-import { HubNavigationParams } from "@instill-ai/toolkit";
 
 export async function generateMetadata() {
   const metadata: Metadata = {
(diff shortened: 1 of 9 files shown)


apps/console/src/middleware.ts
```diff
@@ -1,16 +1,33 @@
+import createMiddleware from "next-easy-middlewares";
```
EiffelFly, line 1: @orangecoloured Could you add the reason why we use this lib?

To solve this kind of issue we can use the conditional middleware https://nextjs.org/docs/app/building-your-application/routing/middleware#conditional-statements To do additional work on the hub related logic. 

This lib's adaption rate worry me, normally I won't put library with such small participant into our codebase

apps/console/src/app/hub/[type]/[list]/page.tsx
Comment (EiffelFly) [inline apps/console/src/app/hub/[type]/[list]/page.tsx:1]: @orangecoloured Hub related route should not be in the ce console, it should only stay in the toolkit library

Comment (EiffelFly) [reply apps/console/src/app/hub/[type]/[list]/page.tsx:1]: You will need to put these related code into console-cloud

EiffelFly (commented):
@orangecoloured Here are some of my thought

PR #1222, opened 2024-06-12
fix(pipelines): change button build to edit

thewbuk:
Because

- Inconsistency in button wording between pages

This commit

- Changes "Build" to "Edit" with icon

INFO:

- Removes stroke from Icon, but all instances use inline stroke colour


EiffelFly (approved):
LGTM

PR #1232, opened 2024-06-17
chore: add icons

orangecoloured:
Because

- needed new icons for credits top up view

This commit

- added them


EiffelFly (approved):
LGTM

PR #1254, opened 2024-06-20
fix: some props of the model trigger payload had snake_case, fixed them to camelCase

orangecoloured:
Because

- The model page still had `snake_case` strings

This commit

- fix `snake_case` to `camelCase`

### DIFF
--- a/packages/toolkit/src/view/model/view-model/ModelOverview.tsx
+++ b/packages/toolkit/src/view/model/view-model/ModelOverview.tsx
@@ -50,6 +50,9 @@ const selector = (store: InstillStore) => ({
 const convertTaskNameToPayloadPropName = (taskName?: ModelTask) =>
   taskName
     ? convertSentenceToCamelCase(
+        // This removes "TASK_" and replaces "_" with a space. The first
+        // argument has and OR operator for matching both substrings. The second
+        // argument is a function with a condition.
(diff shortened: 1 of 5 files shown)


packages/toolkit/src/view/model/view-model/ModelOverview.tsx
```diff
@@ -47,7 +48,11 @@ const selector = (store: InstillStore) => ({
 });
 
 const convertTaskNameToPayloadPropName = (taskName?: ModelTask) =>
-  taskName ? taskName.replace("TASK_", "").toLowerCase() : null;
+  taskName
+    ? convertSentenceToCamelCase(
+        taskName.replace(/TASK_|_/g, (d) => (d === "TASK_" ? "" : " "))
+      )
```
EiffelFly, line 54: @orangecoloured Please leave comment for this kind of code, especially address what is your intention

EiffelFly (commented):
@orangecoloured There is only one thing we need to address in this PR

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
orangecoloured: Done!

EiffelFly (approved):
LGTM

PR #1259, opened 2024-06-24
feat: model view tweaks

orangecoloured:
Because

This commit

### DIFF
--- a/packages/toolkit/src/lib/vdp-sdk/model/types.ts
+++ b/packages/toolkit/src/lib/vdp-sdk/model/types.ts
@@ -127,7 +127,7 @@ export type ModelWatchState = {
   message: string;
 };
 
-export type ModelsWatchState = Record<string, ModelWatchState>;
+export type ModelsWatchState = Record<string, ModelWatchState | null>;
 
 export type ModelTriggerResult = {
   operation: Nullable<
--- a/packages/toolkit/src/lib/vdp-sdk/model/types.ts
+++ b/packages/toolkit/src/lib/vdp-sdk/model/types.ts
@@ -127,7 +127,7 @@ export type ModelWatchState = {
   message: string;
 };
 
-export type ModelsWatchState = Record<string, ModelWatchState | null>;
+export type ModelsWatchState = Record<string, Nullable<ModelWatchState>>;
 
 export type ModelTriggerResult = {
   operation: Nullable<
--- a/packages/toolkit/src/lib/vdp-sdk/model/types.ts
+++ b/packages/toolkit/src/lib/vdp-sdk/model/types.ts
@@ -116,7 +116,7 @@ export type ModelRegion = {
 
 export type ModelVersion = {
   name: string;
-  id: string;
+  version: string;
   digest: string;
   state: ModelState;
   updateTime: string;
--- a/packages/toolkit/src/view/model/ModelsTable.tsx
+++ b/packages/toolkit/src/view/model/ModelsTable.tsx
@@ -70,16 +70,12 @@ export const ModelsTable = (props: ModelsTableProps) => {
       accessorKey: "state",
       header: () => <div className="min-w-[100px] text-center">Status</div>,
       cell: ({ row }) => {
-        const name: string = row.original.name;
+        const name = row.original.name;
 
         return (
           <div className="grid justify-items-center">
             <StateLabel
-              state={
-                (modelsWatchState.isSuccess &&
-                  modelsWatchState.data[name]?.state) ||
-                "STATE_UNSPECIFIED"
-              }
+              state={modelsWatchState.data?.[name]?.state}
             />
           </div>
         );
--- a/packages/toolkit/src/view/model/ModelsTable.tsx
+++ b/packages/toolkit/src/view/model/ModelsTable.tsx
@@ -75,11 +75,7 @@ export const ModelsTable = (props: ModelsTableProps) => {
         return (
           <div className="grid justify-items-center">
             <StateLabel
-              state={
-                modelsWatchState.isSuccess
-                  ? modelsWatchState.data[name]?.state
-                  : "STATE_UNSPECIFIED"
-              }
+              state={(modelsWatchState.isSuccess && modelsWatchState.data[name]?.state) || "STATE_UNSPECIFIED"}
             />
           </div>
         );
--- a/packages/toolkit/src/view/model/ModelsTable.tsx
+++ b/packages/toolkit/src/view/model/ModelsTable.tsx
@@ -75,7 +75,11 @@ export const ModelsTable = (props: ModelsTableProps) => {
         return (
           <div className="grid justify-items-center">
             <StateLabel
-              state={(modelsWatchState.isSuccess && modelsWatchState.data[name]?.state) || "STATE_UNSPECIFIED"}
+              state={
+                (modelsWatchState.isSuccess &&
+                  modelsWatchState.data[name]?.state) ||
+                "STATE_UNSPECIFIED"
+              }
             />
           </div>
         );
--- a/packages/toolkit/src/view/model/ModelsTable.tsx
+++ b/packages/toolkit/src/view/model/ModelsTable.tsx
@@ -74,9 +74,7 @@ export const ModelsTable = (props: ModelsTableProps) => {
 
         return (
           <div className="grid justify-items-center">
-            <StateLabel
-              state={modelsWatchState.data?.[name]?.state}
-            />
+            <StateLabel state={modelsWatchState.data?.[name]?.state} />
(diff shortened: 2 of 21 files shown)


packages/toolkit/src/lib/vdp-sdk/model/types.ts
```diff
@@ -127,7 +127,7 @@ export type ModelWatchState = {
   message: string;
 };
 
-export type ModelsWatchState = Record<string, ModelWatchState>;
+export type ModelsWatchState = Record<string, ModelWatchState | null>;
```
EiffelFly, line 130: @orangecoloured We can use Nullable here

packages/toolkit/src/view/model/ModelsTable.tsx
```diff
@@ -76,9 +76,9 @@ export const ModelsTable = (props: ModelsTableProps) => {
           <div className="grid justify-items-center">
             <StateLabel
               state={
-                modelsWatchState.isSuccess
-                  ? modelsWatchState.data[name]?.state
-                  : "STATE_UNSPECIFIED"
+                (modelsWatchState.isSuccess &&
+                  modelsWatchState.data[name]?.state) ||
+                "STATE_UNSPECIFIED"
```
EiffelFly, line 81: @orangecoloured Please use `ternary operator` instead of short-cutting in this scenario (We only use short-cutting when we want to set default value)

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
```diff
@@ -115,98 +126,106 @@ export const ModelOverview = ({ model, modelState }: ModelOverviewProps) => {
...
-    existingTriggerState,
-    toast,
-    existingTriggerToastShowed,
-    dismiss,
-    isModelRunInProgress,
-    model,
-    modelRunResult,
-    inputFromExistingResult,
-  ]);
+  }, [existingModelTriggerResult]);
```
EiffelFly, line 187: @orangecoloured reactQuery has lots of state in its return, you need to be very specific about what you are going to subscribe with this useEffect hook. 

For example, in this hook, the only two things you need are `existingModelTriggerResult.isSuccess` and `existingModelTriggerResult.data`

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
```diff
@@ -115,98 +126,106 @@ export const ModelOverview = ({ model, modelState }: ModelOverviewProps) => {
...
-    if (!existingModelTriggerResult.data?.operation || !model) {
-      return;
-    }
+  const pollForResponse = React.useCallback(async () => {
+    await queryClient.invalidateQueries({
+      queryKey: ["models", "operation", model?.name],
+    });
 
+    forceUpdate();
+  }, [forceUpdate]);
```
EiffelFly, line 139: @orangecoloured This is a bit hacky to me, normally when the cached data changed, all the related subscribed components and hook should change too. I am wondering what is the issue here?

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
```diff
@@ -248,40 +267,15 @@ export const ModelOverview = ({ model, modelState }: ModelOverviewProps) => {
...
-      });
-
       if (amplitudeIsInit) {
         sendAmplitudeData("trigger_model");
       }
 
+      currentOperationIdPollingData.current = {
+        ...defaultCurrentOperationIdPollingData,
+        name: data.operation.name,
+      };
```
EiffelFly, line 277: @orangecoloured I am curious why you choose to use ref for this data. If it's due to the infinite re-rendering then it might be the wrongly usage of reactQuery object as deps issue I mentioned above

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
EiffelFly, line 1: @orangecoloured Please use 

`import * as React from "react"` to import react

EiffelFly (commented):
I left some question in the comments, please check them

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
orangecoloured: I invalidate the response here and force the hooks to execute again, so the query hook actually tries to fetch the new data and not get it from cache. Basically I'm trying to mimic executing a simple fetch with a timeout.

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
orangecoloured: ref is updated out of the render loop, like a normal object, and I need this data to be updated as soon as I update it and not on the next re-render. By the time next re-render occurs all the hooks already have the fresh data.

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
EiffelFly: This is where I am a bit lost, please shed some light on it

```ts
await queryClient.invalidateQueries({
      queryKey: ["models", "operation", model?.name],
    });
```

This will invalidate the operation query. And ideally, the updated data will trigger the re-render so you can get the up-to-date data. 

Then you trigger the `forceUpdate` function which make the whole component re-render. If the react-query state change already trigger the re-render. Why you need to jump in the life-cycle and force it to render?

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
EiffelFly: @orangecoloured 

> and I need this data to be updated as soon as I update it and not on the next re-render

I think this is the place I am lost. What is the scenario that you need to stay into the react life-cycle and do something in between

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
orangecoloured: Do you mean that calling `invalidateQueries` triggers the hook to fetch the data? I thought it just invalidates the cached response so next time the hook is executed it makes a new fetch instead of giving the cached result.

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
orangecoloured: One of the main reasons - so I don't have to deal with stale data. I don't want to think about the order of the renders. I want to be sure that I can really on this temp storage that keeps the currently active operation id. I want to be sure that at this moment it keeps what I think it should keep. I don't want to also keep in mind the react render cycle.

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
EiffelFly: > Waiting for queries to become stale before they are fetched again doesn't always work, especially when you know for a fact that a query's data is out of date because of something the user has done. For that purpose, the QueryClient has an invalidateQueries method that lets you intelligently mark queries as stale and potentially refetch them too!

https://tanstack.com/query/v4/docs/framework/react/guides/query-invalidation

According to the docs, in some scenario `invalidateQueries` will trigger refetch. But some scenario don't, it depends on whether your query is already rendered or not.

If you want a much straight, you can use refetch function (At the end of this page https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)

CleanShot 2024-06-24 at 20 22 48

orangecoloured: Fixed

packages/toolkit/src/view/model/view-model/ModelOverview.tsx
EiffelFly: I think it's a valid act for me, please leave the comment in the codebase. 

But I would like to address that, normally I don't think leaving react life cycle is a good thing in the long term, people will find it's a bit hard to read through the code if we use ref to dodge the natural life-cycle of React. 

In the end, the lifecycle of the react is the first and formost item to be considered when written react code

orangecoloured: Agree on React lifecycle. But with a complex logic like this, hooks are hard to deal with. Implementing this with a class would be much easier and cleaner.

And I think what could help us here is some kind of a long-polling request or [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).

EiffelFly: Yes, I would say after we have proper streaming among our operation endpoint, this will be much easier to implement, let's hold like this way first

EiffelFly (approved):
LGTM!

PR #1268, opened 2024-07-01
fix: fixed private tag in the pipeline/model header

orangecoloured:
Because

- the styles of the tag were different

This commit

- unified the styles and updated the logic to show only "Private" tag


EiffelFly (approved):
LGTM

PR #1269, opened 2024-07-01
fix: Explore link inactive/hover state

orangecoloured:
Because

- the Explore link lacked the inactive and hover states

This commit

- added Explore link inactive and hover states


EiffelFly (approved):
LGTM

EiffelFly (approved):
LGTM

PR #1279, opened 2024-07-02
feat: update image upload inputs according to the designs

orangecoloured:
Because

- we had our image upload inputs with an outdated UI

This commit

- updated the UX for the image upload inputs


EiffelFly (approved):
@orangecoloured LGTM

This is a great implementation, thanks you 👍

PR #1289, opened 2024-07-11
feat: Pipeline UI alignment with models

orangecoloured:
Because

- The pipelines UI was outdated

This commit

- Updated the pipeline related UI in alignment with models


packages/toolkit/src/lib/react-query-service/metric/usePipelineTriggerRecords.ts
```diff
@@ -19,8 +19,18 @@ export function usePipelineTriggerRecords({
   retry?: false | number;
   filterId: Nullable<string>;
 }) {
+  const queryKey = ["metrics", "pipelines", "triggers"];
+
+  if (filter) {
+    queryKey.push(filter);
+  }
```
EiffelFly, line 26: Since this filter has timeStamp, it will constantly changing. 

- First render: 2024T....-00:09:00 -> Request -> Trigger second render
- Second render: 2024T....-00:09:02

It will form a endless render, this is the reason I remove it in previous fix and use filterId

packages/toolkit/src/lib/vdp-sdk/pipeline/actions.ts
```diff
@@ -173,7 +171,6 @@ export async function triggerUserPipelineReleaseAction({
       payload,
       {
         headers: {
-          "instill-return-traces": returnTraces ? "true" : "false",
```
EiffelFly, line 176: @orangecoloured What is the reason removing this field? We need this to gain the traces

EiffelFly (commented):
@orangecoloured I have some questions, other than that looks good

packages/toolkit/src/lib/react-query-service/metric/usePipelineTriggerRecords.ts
orangecoloured: Makes sense. Fixed my code.

packages/toolkit/src/lib/vdp-sdk/pipeline/actions.ts
orangecoloured: The code does not use that prop.
```
"Access-Control-Allow-Headers": "instill-return-traces, instill-share-code, Instill-Requester-Uid",
```
It always adds that header.

EiffelFly: Oh... yes you are right, nvm, all these action is revamped in the sdk PR

PR #1298, opened 2024-07-15
fix: minor tweaks for the readme editor

orangecoloured:
Because

- there were minor readme editor UI differences with the sigma design

This commit

- updated the readme UI


EiffelFly (approved):
LGTM

PR #1309, opened 2024-07-16
fix: model public requests

orangecoloured:
An update for the model related requests to allow for unauthorized users to get model, model state and model versions.


EiffelFly (approved):
LGTM!

PR #1326, opened 2024-07-28
feat: move restapi component into generic type

donch1989:
Because

- We've reclassified the `restapi` component as a `generic` component.

This commit

- Update the `restapi` component to be categorized as a `generic` component.
- Adds icons for new components.


EiffelFly (approved):
LGTM

PR #1337, opened 2024-08-01
fix: make multiple graphics inputs use cover object fit instead of contain

orangecoloured:
Because

- multiple graphics elements inputs had white backgrounds , because of the `object-fit: contain` css prop

This commit

- made them use `object-fit: cover`


EiffelFly (approved):
LGTM

PR #1343, opened 2024-08-04
fix: fix domain in catolog api 

thewbuk:
Because

- Change single to double quotes in the CURL command
- Some UI adjustments to align with Modals/Pipelines
- Do not display credit notifications on local
- Use correct domain in catalog API commands 
- Change item status to lowercase 
- Adjust sidebar to to hide all other tabs if in "My Catalogs" tab


packages/toolkit/src/view/knowledge/KnowledgeBaseView.tsx
```diff
@@ -56,6 +57,7 @@ export const KnowledgeBaseView = (props: KnowledgeBaseViewProps) => {
   const { accessToken, enabledQuery, selectedNamespace } = useInstillStore(
     useShallow(selector),
   );
+  const isLocalEnvironment = env("NEXT_PUBLIC_API_GATEWAY_URL");
```
EiffelFly, line 60: @thewbuk This is not safe, you can use env("NEXT_PUBLIC_APP_ENV") === "CE" to identify ce environment

EiffelFly (approved):
LGTM

PR #1346, opened 2024-08-05
fix: add pagination to Catalog

thewbuk:
Because

- Add load all resources for Catalog and Chunks tabs

### DIFF
--- a/packages/toolkit/src/view/knowledge/components/tabs/ChunkTab.tsx
+++ b/packages/toolkit/src/view/knowledge/components/tabs/ChunkTab.tsx
@@ -1,8 +1,11 @@
 "use client";
 
 import React from "react";
+import { Nullable } from "instill-sdk";
+
 import { Button, Separator, Skeleton } from "@instill-ai/design-system";
 
+import { EmptyView } from "../../../../components";
 import {
   InstillStore,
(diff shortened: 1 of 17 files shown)


packages/toolkit/src/view/knowledge/components/tabs/ChunkTab.tsx
```diff
@@ -50,23 +55,25 @@ export const ChunkTab = ({ knowledgeBase, onGoToUpload }: ChunkTabProps) => {
...
   const files = React.useMemo(() => {
-    return (
-      allFiles?.filter(
-        (file) => file.processStatus !== "FILE_PROCESS_STATUS_FAILED",
-      ) || []
+    return (filesData?.files || []).filter(
+      (file) => file.processStatus !== "FILE_PROCESS_STATUS_FAILED",
     );
-  }, [allFiles]);
+  }, [filesData]);
```
orangecoloured, line 74: Could you update this yo use `isSuccess` from `useListKnowledgeBaseFiles`?
And then we should use `[knowledgeBaseFiles.isSuccess, knowledgeBaseFiles.data]` as deps.

orangecoloured (changes requested):

PR #1367, opened 2024-08-12
feat: show private/public labels for model cards and on model page

orangecoloured:
Because

- we need to show the private/public model status

This commit

- uncomment the labels code


EiffelFly (approved):
LGTM

PR #1370, opened 2024-08-13
fix(catalog): fix link on Team plan 

thewbuk:
Because

- Fixes link on Team Plan

### DIFF
--- a/packages/toolkit/src/view/knowledge/components/notifications/InsufficientStorageBanner.tsx
+++ b/packages/toolkit/src/view/knowledge/components/notifications/InsufficientStorageBanner.tsx
@@ -1,17 +1,34 @@
 "use client";
 
 import { useRouter } from "next/navigation";
-
 import { Button, Icons, LinkButton } from "@instill-ai/design-system";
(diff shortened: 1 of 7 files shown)


packages/toolkit/src/view/knowledge/components/notifications/InsufficientStorageBanner.tsx
```diff
@@ -1,18 +1,38 @@
...
+  namespaceType,
+  selectedNamespace,
 }: InsufficientStorageBannerProps) => {
   const router = useRouter();
 
+  const getUpgradeLink = () => {
+    if (plan === "PLAN_FREE" && namespaceType === "user") {
+      return "/subscribe";
+    } else if (plan === "PLAN_FREE" && namespaceType === "organization") {
+      return `/${selectedNamespace}/organization-settings/billing/subscriptions/plan`;
```
EiffelFly, line 27: @thewbuk This is not correct. 

The namespace is id not the name, so you need to check what is its type using the useNamespaceType

PR #1371, opened 2024-08-13
fix(catalog): fix rename knowledge base to catalog

thewbuk:
Because

- naming convention changed from knowledge base to catalog


packages/toolkit/src/lib/react-query-service/catalog/types.ts
```diff
@@ -27,7 +27,7 @@ export type FileSnippet = {
   score: number;
 };
 
-export interface KnowledgeFile {
+export interface CatalogFile {
   fileUid: string;
   name: string;
   type: string;
```
EiffelFly, line 33: @thewbuk We use type instead of interface

EiffelFly (commented):
@thewbuk Other than this I think we don't need much change in this Pr

EiffelFly (approved):
LGTM

PR #1390, opened 2024-08-26
fix: fetch specific version operation

orangecoloured:
Because

- On the model page there was no way to fetch specific version last run result

This commit

- Now there is


EiffelFly (commented):
@orangecoloured Some minor change

EiffelFly (approved):
LGTM

PR #1393, opened 2024-08-27
feat: removed Test button from pipeline versions table

orangecoloured:
Because

- no need to the test button when we have the versions dropdown

This commit

- removed the test button from the pipeline versions table


EiffelFly (approved):
LGTM

PR #1411, opened 2024-08-29
feat: Integration page

orangecoloured:
This PR has no "Test" connection feature, because it's not yet available on BE.


packages/sdk/src/core/integration/IntegrationClient.ts
```diff
@@ -0,0 +1,325 @@
...
+      return Promise.resolve(data.connection);
+    } catch (error) {
+      return Promise.reject(error);
+    }
+  }
+
+  /**
+   * Updates an integration connection
+   */
+  async updateIntegrationConnection(props: UpdateIntegrationConnectionRequest) {
```
EiffelFly, line 277: @orangecoloured Let's directly deconstruct the props here since we didn't use it as a whole afterward

packages/toolkit/src/lib/react-query-service/mgmt/useIntegrationConnection.ts
```diff
@@ -0,0 +1,45 @@
...
+    queryKey: ["integration-connection", connectionId, namespaceId, view],
+    queryFn: async () => {
+      if (!connectionId) {
+        return Promise.reject(new Error("integrationId not provided"));
+      }
+
+      if (!namespaceId) {
+        return Promise.reject(new Error("namespaceId not provided"));
+      }
+
```
EiffelFly, line 31: @orangecoloured We need to check accessToken here since this is a query for private resource

packages/toolkit/src/view/settings/integrations/AvailableIntegration.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+const METHOD_TITLE: Record<IntegrationMethod, string> = {
+  METHOD_DICTIONARY: "Dictionary",
+  METHOD_OAUTH: "oAuth",
+};
+
+export type AvailableIntegrationProps = {
+  integration: Integration;
+  accessToken: Nullable<string>;
+  enableQuery: boolean;
```
EiffelFly, line 18: @orangecoloured To avoid props drilling we can use the useInstillStore to get these two value

packages/toolkit/src/view/settings/integrations/ConnectionForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+    let parsedData;
+
+    try {
+      parsedData = ValidatorSchema.parse(payload);
+    } catch (err) {
+      if (err instanceof z.ZodError) {
+        console.log(err);
+      }
+
+      return;
```
EiffelFly, line 85: @orangecoloured We need to do something about this error since it will block user action.

EiffelFly (commented):
@orangecoloured The code is good, just some minor things to address.

1. There are still many components which use hook but don't have `use client`
2. We can use `useInstillStore` to get `accessToken` and `enabledQuery` to avoid props drilling
3. I will suggest that for the react-query queryKey, write a function like `getIntgrationQueryKey` for it to be resuable

packages/toolkit/src/view/settings/integrations/ConnectionForm.tsx
orangecoloured: I added a toast with the error message. It's not blocking the UI. You can close the modal.

orangecoloured:
@EiffelFly What do you mean by reusable `getIntegrationQueryKey`? Those keys are different for every request and those are just string arrays. I don't understand what can be automated here.

EiffelFly:
@orangecoloured 

We have some [references](https://github.com/instill-ai/console/blob/a158fa0c8584415be1a495eee604ca2d3a6e0099/packages/toolkit/src/lib/react-query-service/pipeline/use-namespace-pipeline/server.ts#L38) in the codebase

The reason we try to use this pattern is double-fold

- From time to time, there might be chance that we need to do something about the cache after the mutation, and we need the queryKey, but without providing a function to generate the pattern, developer might get lost or simply use the wrong queryKey
- It's reusable and traceable

orangecoloured:
@EiffelFly alright
so do you mean to have a unique query key generator per endpoint or a general integrations generator?
something like
```
["integrations", "list", "infinite", ...]
["integrations", "connections", "list", "infinite", ...]
```

EiffelFly:
@orangecoloured Yes, what I mean is a "unique query key generator" per endpoint. We don't need a universal one yet

packages/toolkit/src/view/settings/integrations/Integrations.tsx
```diff
@@ -0,0 +1,205 @@
...
+      ) {
+        integrationConnections.fetchNextPage();
+      }
+    }
+  }, [
+    availableIntegrations.isSuccess,
+    availableIntegrations.data,
+    integrationConnections.isSuccess,
+    integrationConnections.data,
+  ]);
```
EiffelFly, line 71: @orangecoloured We can use "query.isSuccess" to check whether data is there, and I see you already include it in the useEffect deps

packages/toolkit/src/view/settings/integrations/Integrations.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const availableIntegrationList = React.useMemo(() => {
+    return (
+      availableIntegrations.data?.pages.reduce(
+        (acc: Integration[], page) => acc.concat(page.integrations),
+        [],
+      ) ||
+      //.sort((a, b) => (a.title < b.title ? -1 : a.title > b.title ? 1 : 0))
+      []
+    );
```
EiffelFly, line 81: @orangecoloured Try not to use reduce to concat the array, simply use for loop to make it easier to read

packages/toolkit/src/view/settings/integrations/PipelineList.tsx
```diff
@@ -0,0 +1,78 @@
...
+    filter: null,
+  });
+
+  const pipelineIds = React.useMemo(() => {
+    return (
+      relatedPipelineIds.data?.pages.reduce(
+        (acc: string[], page) => acc.concat(page.pipelineIds),
+        [],
+      ) || []
+    );
```
EiffelFly, line 42: @orangecoloured Let's use for-loop instead of reduce here

packages/toolkit/src/view/settings/integrations/PipelineList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [relatedPipelineIds.isSuccess, relatedPipelineIds.data]);
+
+  React.useEffect(() => {
+    if (relatedPipelineIds.data) {
```
EiffelFly, line 46: @orangecoloured Let's use isSuccess to guard whether data is there

packages/toolkit/src/view/settings/integrations/PipelineList.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from "react";
```
EiffelFly, line 1: @orangecoloured This component need to have "use client"

EiffelFly (commented):
@orangecoloured Some minor change and then we can merge this pR

PR #1428, opened 2024-09-04
fix: update payload generator for any files

orangecoloured:
Because

- [Slack message](https://instill-ai.slack.com/archives/C07BKQKUW1X/p1725455445171839)

This commit

- adds a case for any file for the pipeline payload generator

### DIFF
--- a/packages/toolkit/src/constant/pipeline.ts
+++ b/packages/toolkit/src/constant/pipeline.ts
@@ -76,6 +76,10 @@ export const generateInputsPayload = (fields: PipelineVariableFieldMap) => {
           ];
           break;
         }
+        case "*/*": {
+          input[key] = "Please put your file base64 encoded string";
+          break;
+        }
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/constant/pipeline.ts
```diff
@@ -76,6 +76,10 @@ export const generateInputsPayload = (fields: PipelineVariableFieldMap) => {
           ];
           break;
         }
+        case "*/*": {
+          input[key] = "Please put your file base64 encoded string";
+          break;
+        }
```
EiffelFly, line 82:
```suggestion
        case "file":
        case "*/*": {
          input[key] = "Please put your file base64 encoded string";
          break;
        }
```

@orangecoloured Please help me add this condition too. Since backend will soon change the case and we can adapt this right now

EiffelFly (approved):
LGTM

PR #1465, opened 2024-09-13
fix: fix model playground polling

orangecoloured:


EiffelFly (approved):
LGTM!! @orangecoloured Thanks for adding the comment

PR #1466, opened 2024-09-15
feat(catalog): feat add warning when cloning to full namespace 

thewbuk:
Because

- Add warning when trying to clone to full namespace


packages/toolkit/src/view/catalog/components/CloneCatalogDialog.tsx
```diff
@@ -78,9 +101,55 @@ export const CloneCatalogDialog = ({
...
+    enabled: enabledQuery && namespaceType === "organization",
+  });
+
+  const subscriptionInfo = React.useMemo(() => {
+    return getSubscriptionInfo(
+      namespaceType,
+      userSub.data || null,
+      orgSub.data || null,
+    );
+  }, [selectedNamespace, userSub.data, orgSub.data, namespaceType]);
```
EiffelFly, line 146: @thewbuk This deps has an unnecessary deps `selectedNamespace`, If your hook didn't work without this variables, then you might need to rethink a bit this useMemo usage

EiffelFly (commented):
@thewbuk Some minor needed changes, please take a look

packages/toolkit/src/view/catalog/components/notifications/CatalogLimitNotification.tsx
```diff
@@ -0,0 +1,46 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+};
+
+export const CatalogLimitNotification = ({
+  isOpen,
+  onClose,
+}: CatalogLimitNotificationProps) => {
+  return (
+    <Dialog.Root open={isOpen} onOpenChange={onClose}>
```
EiffelFly, line 13: @thewbuk This naming is not straight forward, since onOpenChange will get triggered when dialog is opened.

You can directly name it as `setIsOpen` or `onOpenChange`

EiffelFly (commented):
@thewbuk Just discovered a minor issue

And please cast `use client` on all the component where react hook is used

EiffelFly (approved):
LGTM

PR #1481, opened 2024-09-18
fix: add requester ID to the model operation request

orangecoloured:
The model operation endpoint required the Instill requester id to be present in headers.


EiffelFly (approved):
LGTM

PR #1494, opened 2024-09-21
feat: hint transformed id on model and pipeline creation

EiffelFly:
Because

- hint transformed id on model and pipeline creation
CleanShot 2024-09-02 at 11 05 10 2@2x

This commit

- hint transformed id on model and pipeline creation

### DIFF
--- a/packages/toolkit/src/view/recipe-editor/popovers/PipelineNamePopover.tsx
+++ b/packages/toolkit/src/view/recipe-editor/popovers/PipelineNamePopover.tsx
@@ -30,6 +30,7 @@ import {
   useShallow,
   useUpdateNamespacePipeline,
 } from "../../../lib";
+import { formatResourceId } from "../../../server";
 import { EditorButtonTooltipWrapper } from "../EditorButtonTooltipWrapper";
 
 const pipelineNameSchema = z.object({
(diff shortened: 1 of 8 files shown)


packages/toolkit/src/view/recipe-editor/popovers/PipelineNamePopover.tsx
```diff
@@ -72,13 +77,16 @@ export const PipelineNamePopover = ({
     });
   }, [routeInfo.isSuccess, routeInfo.data.resourceId, form, pipelineIsPublic]);
 
+  const formattedPipelineId = formatResourceId(form.watch("id"), "p");
```
orangecoloured, line 80: Should we move `p`, `m`, `c` etc. to a separate constants file?

orangecoloured (changes requested):

packages/toolkit/src/view/recipe-editor/popovers/PipelineNamePopover.tsx
EiffelFly: Oh that is a good suggestion, I will do that 🤝

thewbuk (dismissed):
LGTM

orangecoloured:
@EiffelFly also, how about we do `p_bla-bla`? With a `_` separating the prefix? I mean if t's a prefix, makes sense to have it easily visible.

EiffelFly:
@orangecoloured 

> also, how about we do p_bla-bla? With a _ separating the prefix? I mean if t's a prefix, makes sense to have it easily visible.

Oh yes, that is a good suggestion too. But because we don't allow underscore in some place, I think we can make it like

`p-`, `c-` and `m-`

PR #1501, opened 2024-09-24
chore: add Collection component

donch1989:
Because

- We have a new Collection component

This commit

- Adds icon
- Adds component ID into component list


EiffelFly (approved):
LGTM

PR #1505, opened 2024-09-24
fix(explore): fix capitalise some tags

thewbuk:
Because

- Capitalise Featured tag and Private/Public label 
- Change Triggered to runs
- Change Runs icon in models


EiffelFly (approved):
LGTM

PR #1512, opened 2024-10-01
feat(assistant): feat add assistant navbar 

thewbuk:
Because

- Add navbar to assistant navbar


packages/toolkit/src/components/top-bar/NavLinks.tsx
```diff
@@ -128,19 +134,23 @@ export const NavLinks = ({ isExploreRoute }: { isExploreRoute?: boolean }) => {
...
-              key={pathname}
-              pathname={pathname}
-              Icon={Icon}
-              title={title}
-              isExploreRoute={isExploreRoute}
-              strict={true}
-            />
-          ))
+        ? navLinkItems
+            .filter((item) => isCloud || item.pathname !== "applications")
```
EiffelFly, line 143: @thewbuk This logic is kind of hard to read, let's don't use one-liner here, instead

if (isCloud){
  return true
}

return item.pathname !== "application"

EiffelFly (commented):
@thewbuk One small logic thing

EiffelFly (approved):
LGTM

PR #1532, opened 2024-10-12
feat(application): feat export functions to helpers 

thewbuk:
Because

- Migrate additional functions to helpers


EiffelFly (approved):
LGTM

PR #1535, opened 2024-10-14
fix(application): fix streaming

thewbuk:
Because

- Add streaming to chat

### DIFF
--- a/packages/sdk/src/application/ApplicationClient.ts
+++ b/packages/sdk/src/application/ApplicationClient.ts
@@ -462,7 +462,6 @@ export class ApplicationClient extends APIResource {
           "Content-Type": "application/json",
           ...additionalHeaders,
           Accept: "text/event-stream",
-          Authorization: `Bearer ${requesterUid}`,
         },
         body: JSON.stringify(payload),
       });
--- a/packages/sdk/src/application/ApplicationClient.ts
+++ b/packages/sdk/src/application/ApplicationClient.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import { getInstillAdditionalHeaders, getQueryString } from "../helper";
 import { APIResource } from "../main/resource";
 import {
--- a/packages/sdk/src/application/ApplicationClient.ts
+++ b/packages/sdk/src/application/ApplicationClient.ts
@@ -3,6 +3,7 @@ import { APIResource } from "../main/resource";
 import {
   Application,
   ChatRequest,
+  ChatResponse,
   Conversation,
   CreateApplicationRequest,
   CreateApplicationResponse,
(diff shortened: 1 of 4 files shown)


packages/sdk/src/application/ApplicationClient.ts
```diff
@@ -438,31 +437,55 @@ export class ApplicationClient extends APIResource {
...
+
+      if (stream) {
+        return response.body as ReadableStream<Uint8Array>;
+      } else {
+        return response.json();
+      }
     } catch (error) {
-      return Promise.reject(error);
+      console.error("Fetch failed:", error);
+      throw error;
```
EiffelFly, line 488: @thewbuk If you are using our class inherited method, it already handle streaming for you, all you need to do is directly return the response you get like this https://github.com/instill-ai/console/blob/109ff456d697af0f65f4c2fee5464554c33657cf/packages/sdk/src/vdp/trigger/TriggerClient.ts#L129

packages/sdk/src/application/ApplicationClient.ts
```diff
@@ -438,31 +437,55 @@ export class ApplicationClient extends APIResource {
...
-  async chat(props: ChatRequest): Promise<ChatResponse> {
-    const { ownerId, appId, ...payload } = props;
+  async chat(
+    props: ChatRequest,
+    stream: boolean = true,
+  ): Promise<ReadableStream<Uint8Array>> {
+    const { ownerId, appId, requesterUid, accessToken, ...payload } = props;
+
+    if (
+      !accessToken ||
```
EiffelFly, line 453: @thewbuk Access token is completely not necessary, is is initize from the constructor of the ApiClient class. Ask user to input it again is a duplicated behavior

packages/sdk/src/application/ApplicationClient.ts
```diff
@@ -438,31 +437,55 @@ export class ApplicationClient extends APIResource {
...
+      !ownerId ||
+      !appId ||
+      !payload.catalogId ||
+      !payload.conversationUid
+    ) {
+      throw new Error("Required parameters are missing");
```
EiffelFly, line 459: @thewbuk Please imagine yourself is the user, you can't debug this error when all you get is the same message, try to get more hint to the user

packages/sdk/src/application/types.ts
```diff
@@ -176,13 +176,17 @@ export type ChatRequest = {
   catalogId: string;
   conversationUid: string;
   message: string;
+  accessToken: string;
```
EiffelFly, line 179: @thewbuk This is unnecessary

packages/sdk/src/application/ApplicationClient.ts
```diff
@@ -438,31 +438,52 @@ export class ApplicationClient extends APIResource {
...
+    try {
+      const data = await this._client.post(queryString, {
         body: JSON.stringify(payload),
         additionalHeaders,
+        stream,
       });
-      return response;
+
+      if (stream) {
+        return Promise.resolve(data as ChatRequest) as Promise<ChatResponse>;
```
EiffelFly, line 480: @thewbuk This type is wrong, you don't need to wrap two layer around it

packages/sdk/src/application/ApplicationClient.ts
```diff
@@ -438,31 +438,55 @@ export class ApplicationClient extends APIResource {
...
       return Promise.reject(error);
     }
   }
 
-  async chat(props: ChatRequest): Promise<ChatResponse> {
-    const { ownerId, appId, ...payload } = props;
+  async chat(
+    props: ChatRequest,
+    stream: boolean = true,
+  ): Promise<ChatResponse | unknown> {
```
EiffelFly, line 450: @thewbuk SDK can't return `unknown` you need to overload this 

- When stream is enable then it's any or ReadableStream
- When stream is disable you return the normal one

EiffelFly (approved):
LGTM

PR #1539, opened 2024-10-15
feat: dynamically update model UI state on run

orangecoloured:


packages/toolkit/src/view/model/ModelHubSettingPageMainView.tsx
EiffelFly, line 1: @orangecoloured We have a missing dependency for updateActiveVersionUrl, we either add its or wrap is within useCallback

EiffelFly (commented):
@orangecoloured I see many lint issues in the realm of model, could you help me fix them in this PR?

EiffelFly (dismissed):
Nice comment 👍

PR #1540, opened 2024-10-15
feat: client cookie management functions

orangecoloured:


packages/toolkit/src/lib/clientCookies.ts
EiffelFly, line 1: @orangecoloured Let's re export these two function under the object `ClientCookie` so we can use them together

`ClientCookie.get()` and `ClientCookie.set()`

EiffelFly (approved):
LGTM

PR #1543, opened 2024-10-17
fix: reset playground i/o data on namespace change

orangecoloured:
To test this this toolkit version could be used for cloud `0.107.0-rc.15`

Because

- changing namespace on model playground did not reset the i/o forms data

This commit

- now it do reset it

https://github.com/user-attachments/assets/f058e1ed-804f-4d53-bcf9-03286eea226a

### DIFF
--- a/packages/toolkit/src/view/model/view-model/ModelPlayground.tsx
+++ b/packages/toolkit/src/view/model/view-model/ModelPlayground.tsx
@@ -156,6 +156,9 @@ export const ModelPlayground = ({
     },
   );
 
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/view/model/view-model/ModelPlayground.tsx
EiffelFly, line 1: @orangecoloured line 285

```
React Hook useEffect has missing dependencies: 'pollForResponse' and 'toast'. Either include them or remove the dependency array
```

EiffelFly (approved):
LGTM

PR #1546, opened 2024-10-18
feat: put back image input preview, clean up

orangecoloured:
Because

- We want to have previews for the image input fields

This commit

- I put it back
- Also removed the image to image task, it's not supported by BE


EiffelFly (approved):
LGTM

PR #1564, opened 2024-10-25
feat: replace vdp-sdk/organization with instill-sdk

EiffelFly:
Because

- vdp-sdk is serverly out-of-date, we want to replace it with our new instill-sdk

This commit

- replace vdp-sdk with instill-sdk


packages/sdk/src/helper/UndefinedToNull.ts
```diff
@@ -0,0 +1,7 @@
+export type UnionUndefinedToNull<T> = T extends undefined ? null : T;
+
+export type UndefinedToNull<T> = {
```
orangecoloured, line 3: What does this type do?

packages/toolkit/src/lib/react-query-service/organization/use-organizations/client.ts
```diff
@@ -0,0 +1,35 @@
...
+}) {
+  let enabledQuery = false;
+
+  if (accessToken && enabled) {
+    enabledQuery = true;
+  }
+
+  const queryKey = getUseOrganizationsQueryKey();
+
+  return useQuery({
```
orangecoloured, line 25: Does it return the orgs based on the access token?

packages/toolkit/src/lib/react-query-service/organization/use-organizations/server.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    return Promise.resolve(organization);
+  } catch (error) {
+    return Promise.reject(error);
+  }
+}
+
+export function getUseOrganizationsQueryKey() {
+  return ["organization"];
```
orangecoloured, line 34: Does it make sense to use plural?

packages/sdk/src/helper/UndefinedToNull.ts
EiffelFly: It will convert all the undefined to null, leave it for future usage

packages/toolkit/src/lib/react-query-service/organization/use-organizations/client.ts
EiffelFly: yes, under the hood BE will use the accessToken to parse the correct user_id

packages/toolkit/src/lib/react-query-service/organization/use-organizations/server.ts
EiffelFly: Oh no... I need to change it

orangecoloured (dismissed):
Mostly looks like the org endpoints got an updated. Otherwise pure code style stuff.

PR #1566, opened 2024-10-25
feat: replace vdp-sdk/model with instill-sdk

EiffelFly:
Because

- replace vdp-sdk/model with instill-sdk
- Adapt namespace structure (users/uid -> namespaces/nid)

This commit

- replace vdp-sdk/model with instill-sdk


packages/toolkit/src/lib/react-query-service/model/use-namespace-models/server.ts
```diff
@@ -0,0 +1,69 @@
...
+      enablePagination: false,
+    });
+
+    return Promise.resolve(models);
+  } catch (err) {
+    return Promise.reject(err);
+  }
+}
+
+export function prefetchNamespaceModels({
```
orangecoloured, line 41: When is it used?

orangecoloured (approved):

packages/toolkit/src/lib/react-query-service/model/use-namespace-models/server.ts
EiffelFly: Currently no, but if we really want to do so SSR prefetch stuff, this will serves the need

PR #1572, opened 2024-10-29
feat(catalog): migrate API calls to sdk CatalogClient

thewbuk:
Because

- migrate catalog to new sdk


packages/toolkit/src/lib/react-query-service/catalog/useListCatalogFiles.ts
```diff
@@ -1,98 +1,39 @@
...
     queryFn: async () => {
-      if (!accessToken) {
-        throw new Error("accessToken not provided");
-      }
-      if (!namespaceId) {
-        throw new Error("namespaceId not provided");
-      }
-      if (!catalogId) {
-        throw new Error("catalogId not provided");
+      if (!namespaceId || !accessToken || !catalogId) {
```
orangecoloured, line 20: Do we want to generalise here this error?
Yeah, I understand that most likely we'll never see it, but still.
In case it throws we won't know which parameter exactly was missing.

packages/sdk/src/catalog/CatalogClient.ts
```diff
@@ -1 +1,371 @@
...
+      });
+
+      const data = await this._client.get<{ sourceFile: FileContent }>(
+        queryString,
+      );
+      return Promise.resolve(data.sourceFile);
+    } catch (error) {
+      return Promise.reject(error);
+    }
+  }
```
EiffelFly, line 251: @thewbuk This should follow the instill.tech API docs name which is `getCatalogSingleSourceOfTruthFile`

https://openapi.instill.tech/reference/artifactpublicservice_getsourcefile

packages/sdk/src/catalog/CatalogClient.ts
```diff
@@ -1 +1,371 @@
...
+
+  async processCatalogFiles(
+    props: ProcessCatalogFilesRequest,
+  ): Promise<ProcessCatalogFilesResponse> {
+    const { fileUids, namespaceUid } = props;
```
EiffelFly, line 256: @thewbuk Please replace `namespaceUid `with `requesterUid`

packages/sdk/src/catalog/CatalogClient.ts
```diff
@@ -1 +1,371 @@
...
+      });
+
+      const data = await this._client.get<{ content: ChunkContent }>(
+        queryString,
+      );
+      return Promise.resolve(data.content);
+    } catch (error) {
+      return Promise.reject(error);
+    }
+  }
```
EiffelFly, line 352: @thewbuk Do we have this one on the docs? I can't find it

thewbuk: I think we had it ages ago. Removed it

EiffelFly:
@thewbuk Please make your PR title as explicit and details as possible

EiffelFly (approved):
LGTM

PR #1573, opened 2024-10-29
fix(catalog): fix remove floating tooltip from card

thewbuk:
Because

- Align with Application card and remove floating tooltip from card


EiffelFly (approved):
LGTM

PR #1574, opened 2024-10-29
feat(catalog): add Requester UID to example code in Ask and Retrieve pages 

thewbuk:
Because

- if user is an organization show requester UID in example code blocks


EiffelFly:
@thewbuk For the PR title you don't need to write the tag again like feat(...): feat ..., you can directly write the description. For me this PR's title is not clear enough, please rethink about it and write a more detail one

EiffelFly (approved):
LGTM, that is much better PR title 👍

PR #1578, opened 2024-11-01
feat: revamp catalog sdk and its react-query

EiffelFly:
Because

@thewbuk Please take a look at this PR, since this is the clean up task for your implementation, I would like to point out several crucial points that might be helpful for your future implementation 

- You need to keep your writing consistent and clean, there have many unnecessary syntax, such as
  - `return Promise.resolve()` -> The async/await already wrap the un-returned function for you
  - You didn't cast `use client` in your react-query, they are hook so can only work on the client side
- You need to follow what BE provided, there are dozens of endpoint didn't provide pagination but you still expose pageSize and pageToken on them such as listNamespaceCatalogs
  - It's pretty clear that you didn't understand the API you are working with well enough

Besides all the above code smell issues, I revamped some structure and make it stay consistent with our other query-hook

- Centralized queryKeyStore for better key cache management
- Separate paginated endpoint and non-paginated endpoint in the sdk
- Retain the response structure in the sdk response, so user need to pour them out like how we did it in the react-query, `res.chunk`

This commit

- revamp catalog sdk and its react-query


thewbuk (approved):
LGTM

PR #1579, opened 2024-11-01
chore(component): rename restapi component to http component

donch1989:
Because

- We've renamed `restapi` component to `http` component

This commit

- rename the icon.


EiffelFly (approved):
LGTM

PR #1584, opened 2024-11-03
feat(catalog): add notification while file is being deleted 

thewbuk:
Because

- Add bottom notification while file is being deleted 
- Remove file upload limit in CE


EiffelFly (approved):
LGTM

PR #1585, opened 2024-11-03
feat(dashboard): Add cost tabs to dashboard and model usage tracking 

thewbuk:
Because

This commit


packages/toolkit/src/view/dashboard/cost/CostTab.tsx
```diff
@@ -0,0 +1,157 @@
...
+      value: "pipeline",
+      label: "Pipeline",
+      icon: <Icons.Pipeline className="h-4 w-4" />,
+    },
+    {
+      value: "model",
+      label: "Model",
+      icon: <Icons.Model className="h-4 w-4" />,
+    },
+  ];
```
EiffelFly, line 98: @thewbuk Try to move this kind of static data outside of component

packages/toolkit/src/lib/react-query-service/metric/useCreditConsumptionChartRecords.ts
```diff
@@ -1,63 +1,46 @@
 "use client";
 
-import type { Nullable } from "instill-sdk";
 import { useQuery } from "@tanstack/react-query";
 
 import { getInstillAPIClient } from "../../sdk-helper";
+import { Nullable } from "../../type";
```
EiffelFly, line 6: @thewbuk Please use Nullable come from instill-sdk

packages/toolkit/src/view/dashboard/UsageSwitch.tsx
EiffelFly, line 1: @thewbuk We have toggelGroup in design-system for this kind of component

EiffelFly (commented):
@thewbuk There are many issues around naming, indentation...etc. please take a look

PR #1587, opened 2024-11-04
fix(catalog): start file upload progress tracking immediately after uploading files 

thewbuk:
Because

- Start progress bar immediately after clicking process files


EiffelFly (approved):
LGTM

PR #1622, opened 2024-11-27
chore(run logging): clarify actors in Run objects

jvallesm:
Because

- The `PipelineRun` and `ModelRun` objects have confusing documentation for
  the requester, runner and owner of a pipeline/model run.

This commit

- Adopts the changes in https://github.com/instill-ai/protobufs/pull/519

### DIFF
--- a/packages/sdk/src/model/types.ts
+++ b/packages/sdk/src/model/types.ts
@@ -194,9 +194,8 @@ export type ListModelRunsRequest = {
 
 export type ModelRun = {
   uid: string;
-  modelUid: string;
   runnerId: string;
-  namespaceId: string;
+  modelNamespaceId: string;
   status: RunStatus;
   source: RunSource;
   totalDuration: number;
--- a/packages/sdk/src/vdp/trigger/types.ts
+++ b/packages/sdk/src/vdp/trigger/types.ts
@@ -11,9 +11,8 @@ import { PipelineTrace } from "../pipeline";
 import { ResourceView, RunSource, RunStatus } from "../types";
 
 export type PipelineRun = {
-  namespaceId: string;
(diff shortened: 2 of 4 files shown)


packages/sdk/src/model/types.ts
```diff
@@ -194,9 +194,8 @@ export type ListModelRunsRequest = {
 
 export type ModelRun = {
   uid: string;
-  modelUid: string;
   runnerId: string;
-  namespaceId: string;
+  modelNamespaceId: string;
```
EiffelFly, line 198: Hi @jvallesm 

I would like to know, will this be the general naming rule of BE after? For example, right now we don't have namespaceId under model object, https://github.com/instill-ai/console/blob/3e14a64adefbb47c66c451a788fc2b99d5a8718d/packages/sdk/src/model/types.ts#L64

- Do we plan to add them in the future?
- If we add it, will we follow the same rule like ModelRuns?

jvallesm: No, this is only for model runs, since _namespace_ there isn't clear enough. Is it the _model_ namespace (owner of the model, yes)? Or the _model run_ namespace (requester of the Run, not necessarily)? I think the requester, runner and namespace fields in the `ModelRun` object are confusing and only become clear when reading the API docs, so if we can add clarity in the names it can improve the dev experience.

For `Model`, namespace is clear and follows the convention, we won't need to change that.

EiffelFly: @jvallesm Thanks for your clarification and this PR looks good to me, thanks for your help 🙏

EiffelFly (approved):
LGTM

jvallesm:
I'll implement the backend part for this today 🏃

PR #1626, opened 2024-11-27
chore(integration): adjust OAuth scope for google-drive component

donch1989:
Because

- we need `drive.file` scope to create/update/delete data in Google Drive

This commit

- adjusts OAuth scope for google-drive component


EiffelFly (approved):
LGTM

PR #1627, opened 2024-11-27
chore(component): add icons for google-drive, perplexity-ai and leadiq

donch1989:
Because

- we have new components: google-drive, perplexity-ai and leadiq

This commit

- adds icons.
- updates component list.


EiffelFly (approved):
LGTM

PR #1636, opened 2024-12-02
chore: support `required` attribute in variable section

donch1989:
Because

- we want users to explicitly define which variables are required.

This commit

- adds support for the required attribute in variables.


EiffelFly (approved):
LGTM

PR #1638, opened 2024-12-02
feat: update add component model UI

orangecoloured:


packages/toolkit/src/view/recipe-editor/commands/ComponentCmdo.tsx
```diff
@@ -777,14 +782,18 @@ export const ComponentCmdo = () => {
...
-            <div className="flex flex-col  w-full h-full">
-              {selectedComponentDefinition &&
-              isComponentDefinition(selectedComponentDefinition) ? (
-                <ScrollArea.Root className="flex shrink-0 mb-auto h-[280px]">
-                  <CommandGroup headingWrapperClassName="px-2" heading="Task">
+            <div className="w-full h-full relative">
+              <div className="absolute top-0 left-0 w-full h-full overflow-hidden">
+                {selectedComponentDefinition &&
+                isComponentDefinition(selectedComponentDefinition) ? (
+                  <ScrollArea.Root className="flex shrink-0 mb-auto h-full [&>*>*>*:last-child]:mb-8">
```
EiffelFly, line 789: @orangecoloured Why we need to use relative and absolute here?

packages/toolkit/src/lib/markdown/MarkdownViewer.tsx
```diff
@@ -16,11 +16,13 @@ export const MarkdownViewer = ({
   markdown,
   skipHtml,
   style,
+  postFix,
 }: {
   className?: string;
   markdown: string;
   skipHtml?: boolean;
   style?: React.CSSProperties;
+  postFix?: React.ReactNode;
```
EiffelFly, line 25: @orangecoloured I would like to know more about this naming, it seems to me like a children but have more meaning to it, can you elaborate a bit?

packages/toolkit/src/view/recipe-editor/commands/ComponentCmdo.tsx
orangecoloured: Because all the parents use height 100% and the list expands too much without a set height. It breaks the layout. So by using an absolute inside the relative I limit its height.

packages/toolkit/src/lib/markdown/MarkdownViewer.tsx
orangecoloured: I had an “append” in mind. And it’s a verb. And we can also prepend. So I thought about prefixes and went with the postfix.

EiffelFly: I think you can name it suffix if that is the opposite of prefix 🤝

EiffelFly (approved):
LGTM

PR #1644, opened 2024-12-09
fix: allow more data types in the variable default field

donch1989:
Because

- the default value in the variable section can represent various data types.

This commit

- expands support for additional data types in the variable default field. Since the default value can be of any data type, we remove the `type` limitation.


EiffelFly (approved):
LGTM

PR #1726, opened 2025-05-20
fix: more file limits removed

orangecoloured:
Addressing this: https://linear.app/instill-ai/issue/INS-7814/bug-catalog-count-and-storage-limit-should-not-be-shown#comment-7d0356ae


EiffelFly (approved):
LGTM

PR #1763, opened 2025-06-13
ci(workflow): update the dockerhub credentials and docker version in ga workflow

Sarthak-instill:
Because

- We have updated the login credentials from drop@instill.tech to drop@instill-ai.com in dockerhub.

This commit

-  update the dockerhub credentials  and docker version in ga workflow.


EiffelFly (approved):
LGTM

PR #1802, opened 2025-07-21
ci(github): trigger instill-core sync after image build

donch1989:
Because

- we want to automatically sync the image hash or tag to the instill-core repository.

This commit

- triggers the instill-core sync after the image is built.


EiffelFly (approved):
LGTM

PR #1808, opened 2025-07-26
feat: support new folder files API

donch1989:
Because

- we want to integrate new folder files API

This commit

- supports new folder files API.
- updates integration-test workflow.


EiffelFly (approved):
LGTM

PR #1814, opened 2025-08-10
chore(github,toolkit,radix,jsx): houseclean  codebase

pinglin:
Because

- the `.github` folder is pretty outdated.
- dashboard complains about `ECharts` import
- Radix UI complains about `ConsoleError: Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.`
- Fix JSX errors:
```
Error: Received `true` for a non-boolean attribute `jsx`.

If you want to write it to the DOM, pass a string instead: jsx="true" or jsx={value.toString()}.
```

This commit

- addresses the above issues.


apps/console/package.json
```diff
@@ -56,12 +56,14 @@
     "js-yaml": "^4.1.0",
     "next": "15.3.3",
     "next-mdx-remote": "^4.1.0",
+    "radix-ui": "^1.4.2",
```
orangecoloured, line 59: Why is this added here?

apps/console/package.json
```diff
@@ -56,12 +56,14 @@
...
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-hook-form": "catalog:",
     "reactflow": "^11.8.3",
     "sharp": "^0.32.6",
     "shiki": "^0.11.1",
+    "styled-jsx": "^5.1.7",
```
orangecoloured, line 66: Same question.

packages/toolkit/src/view/dashboard/activity/ModelsTriggerCountsLineChart.tsx
```diff
@@ -30,12 +29,11 @@ function selectGraph(params: any, myChart: echarts.ECharts): void {
 }
 
 /* eslint-disable-next-line  @typescript-eslint/no-explicit-any */
-function unselectGraph(params: any, myChart: echarts.ECharts): void {
```
orangecoloured, line 33: Could you import only the type and leave the typing in place?
like `import type {.ECharts } from "echarts"`

apps/console/package.json
```diff
@@ -36,9 +36,9 @@
     "@radix-ui/react-dialog": "^1.0.3",
     "@radix-ui/react-label": "^2.0.2",
     "@radix-ui/react-scroll-area": "^1.0.4",
-    "@radix-ui/react-select": "^1.2.1",
+    "@radix-ui/react-select": "^2.2.5",
```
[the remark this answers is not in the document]
pinglin: This is to address `Error: Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.`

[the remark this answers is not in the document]
orangecoloured: Ok. Understood.
Did you test the dependent components? Are they ok with the updated package?

packages/toolkit/src/view/dashboard/activity/ModelsTriggerCountsLineChart.tsx
pinglin: addressed in [commit](https://github.com/instill-ai/console/pull/1814/commits/044cd328fea1c197d2bbf05b64162a6849c3eb39).

apps/console/package.json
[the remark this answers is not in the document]
orangecoloured: No e2e tests. We would just check the occurrences of the dependent components.

orangecoloured (approved):

PR #1819, opened 2025-08-29
fix(toolkit): fix form key missing errors

pinglin:
Because

- on pipeline triggering, the builder reports error `index.mjs:72394 Each child in a list should have a unique "key" prop.`

This commit

- fixes the issue by using index-based keys.

### DIFF
--- a/packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
+++ b/packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
@@ -25,7 +25,7 @@ export const TextsField = (props: TextsFieldProps) => {
           {normalizedTexts && normalizedTexts.length > 0 ? (
             normalizedTexts.map((text, idx) => (
               <MDTextViewer
-                key={`${title}-${idx}`}
+                key={`${title}-${text}-${idx}-field`}
                 text={text}
                 forceFormatted={forceFormatted}
               />
--- a/packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
+++ b/packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
@@ -23,9 +23,9 @@ export const TextsField = (props: TextsFieldProps) => {
       {!hideField ? (
         <div className="flex w-full flex-col flex-wrap gap-2">
           {normalizedTexts && normalizedTexts.length > 0 ? (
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
```diff
@@ -23,9 +23,9 @@ export const TextsField = (props: TextsFieldProps) => {
       {!hideField ? (
         <div className="flex w-full flex-col flex-wrap gap-2">
           {normalizedTexts && normalizedTexts.length > 0 ? (
-            normalizedTexts.map((text) => (
+            normalizedTexts.map((text, idx) => (
               <MDTextViewer
-                key={`${title}-${text}-field`}
+                key={`${title}-${idx}`}
```
orangecoloured, line 28: Could you put back the `text` as well? For better uniqueness.

packages/toolkit/src/lib/use-instill-form/pick/pickComponentOutputFieldsFromInstillFormTree.tsx
```diff
@@ -164,15 +174,17 @@ export function pickComponentOutputFieldsFromInstillFormTree(
...
-          return pickComponentOutputFieldsFromInstillFormTree({
-            ...props,
-            tree: tree.items,
-            data: {
-              [tree.fieldKey as string]: data,
-            },
-          });
-        })}
+        {arrayArrayData.map((data, idx) => (
+          <React.Fragment key={`${tree.path || tree.fieldKey}-${idx}`}>
```
orangecoloured, line 178: I'm not sure you need to use these wrappers here. The returned value from the `pickComponentOutputFieldsFromInstillFormTree` already has the `key` prop.

packages/toolkit/src/lib/use-instill-form/components/component-output/TextsField.tsx
pinglin: Good point.

packages/toolkit/src/lib/use-instill-form/pick/pickComponentOutputFieldsFromInstillFormTree.tsx
pinglin: addressed.

orangecoloured (approved):

PR #1820, opened 2025-08-29
chore(toolkit): add filename in base64-encoded binary payload

pinglin:
Because

- we need to preserve the original filename for base64-encoded binary payload.

This commit

- adds filename metadata in base64-encoded binary payload.

### DIFF
--- a/packages/toolkit/src/view/pipeline-builder/lib/readFileToBinary.ts
+++ b/packages/toolkit/src/view/pipeline-builder/lib/readFileToBinary.ts
@@ -4,7 +4,22 @@ export async function readFileToBinary(file: File) {
 
     reader.onload = (event) => {
       if (event.target) {
-        resolve(event.target.result as string);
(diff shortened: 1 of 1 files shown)


orangecoloured:
@pinglin Just to be sure. Did you test this? And how?

pinglin:
> @pinglin Just to be sure. Did you test this? And how?

@orangecoloured Yes, but it's in my local env with another feature branch in the pipeline-backend repo, if we'd like to see an end-to-end testing result. 

Another way is to print out the `object` in `useUploadAndGetDownloadNamespaceObjectURL.ts` to the console to see if the base64-encoded filename metadata is there or not.

Both approaches require an Instill Core instance spun up in the local and a testing pipeline recipe with a file upload form used.

I will provide a QA screenshot after #1821 is merged and this PR rebased on main.

orangecoloured (dismissed):

PR #1821, opened 2025-08-29
chore(sdk): remove credit-related API

pinglin:
Because

- the credit API has been removed from the CE protobuf and `CreditClient` isn't used in the codebase at all.

This commit

- removes credit-related API.


packages/toolkit/src/components/top-bar/NamespaceSwitch.tsx
```diff
@@ -372,13 +343,7 @@ export const NamespaceSwitch = () => {
...
         setSwitchIsOpen(value);
       }}
-      disabled={
-        env("NEXT_PUBLIC_APP_ENV") === "CLOUD"
-          ? namespacesRemainingCredit.isSuccess
-            ? false
-            : true
-          : false
-      }
+      disabled={false}
```
orangecoloured, line 346: You can just omit passing the disabled, so just remove it.

orangecoloured (changes requested):

packages/toolkit/src/components/top-bar/NamespaceSwitch.tsx
pinglin: addressed.

PR #1822, opened 2025-09-01
fix(console,tookit): fix misleading errors

pinglin:
Because

- the previous logic assumed that data will always be a valid JSON string that can be parsed, but if the API response is null or doesn't contain the expected structure, JSON.parse(data) will fail or return null, causing the error when trying to access `.cookieToken`.
- in the profile setting page, the file input’s id was generated via `uuidv4()` inside a client component. That produces a different value on the server vs client during SSR, which triggers the hydration mismatch you saw in the error diff.
- favicon was outdated.

This commit

- addresses the above issues.
- uses `React.useId()` instead, which is SSR-safe and stable across server and client.
- updates favicon files.


orangecoloured (changes requested):

packages/toolkit/src/lib/react-query-service/mgmt/use-authenticated-user/server.ts
```diff
@@ -17,6 +17,11 @@ export async function fetchAuthenticatedUser({
 
     const user = await client.core.user.getAuthenticatedUser();
 
+    // Sanitize avatar URL to prevent 404 errors
+    if (user.profile?.avatar) {
+      user.profile.avatar = user.profile.avatar.endsWith('avatar') ? undefined : user.profile.avatar;
```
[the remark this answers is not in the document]
pinglin: @orangecoloured Yes, exactly. 

Based on the review comment, I will add a more robust method like:

```javascript
    // Robustly normalize avatar URL to prevent 404s while keeping valid images
    if (user.profile && typeof user.profile.avatar === "string") {
      const original = user.profile.avatar.trim();
      user.profile.avatar = isUsableAvatarUrl(original) ? original : undefined;
    }

...

// Decide whether the provided avatar URL is safe to render in an <img src="..." />
function isUsableAvatarUrl(url: string): boolean {
  if (!url) return false;

  // Data URL images are always safe to render
  if (url.startsWith("data:image/")) return true;

  // Block the known avatar API endpoint (browser won't send auth for <img>)
  if (url.endsWith("/v1beta/users/admin/avatar")) return false;

  // Block general API versioned endpoints as a fallback guard
  const isApiEndpoint = url.includes("/v1beta/") || url.includes("/v1alpha/");
  if (isApiEndpoint) return false;

  const hasImageExtension = /\.(png|jpe?g|gif|webp|svg)(\?.*)?$/i.test(url);
  const isHttp = /^https?:\/\//i.test(url);

  // Accept typical image cases:
  // - explicit image extensions
  // - http(s) URLs that are not API endpoints (CDN or public URLs)
  if (hasImageExtension || (isHttp && !isApiEndpoint)) {
    return true;
  }

  return false;
}
```

pinglin:
Fixes further another bug for the profile image setting layout.

**Before**
CleanShot 2025-09-01 at 20 59 28@2x

**After**
CleanShot 2025-09-01 at 21 00 21@2x

apps/console/src/pages/api/avatar/[userId].ts
Comment (orangecoloured) [inline apps/console/src/pages/api/avatar/[userId].ts:1]: Not sure how this improved things. It simply moved the 404 from the FE to Next.js server function. I think the better way would be to actually fix the BE to return proper links. But ok.

packages/toolkit/src/components/UploadImageFieldWithCrop.tsx
```diff
@@ -57,42 +55,49 @@ export const UploadImageFieldWithCrop = ({
...
+  // Derive initials from form values
+  const displayNameFromForm =
+    (form.getValues?.("profile.displayName") as string | undefined) ?? "";
+  const namespaceIdFromForm =
+    (form.getValues?.("id") as string | undefined) ?? "";
+  const initialsSource =
+    displayNameFromForm.trim() !== ""
+      ? displayNameFromForm
+      : namespaceIdFromForm;
+  const initials = initialsSource
```
orangecoloured, line 97: I think we have `generateFallbackName` for this.

packages/toolkit/src/lib/react-query-service/mgmt/use-authenticated-user/server.ts
```diff
@@ -17,12 +17,58 @@ export async function fetchAuthenticatedUser({
...
     return Promise.reject(error);
   }
 }
 
+// Normalize avatar URL:
+// - data URLs: returned as-is
+// - versioned API endpoints /v1|v1alpha|v1beta/users/{id}/avatar -> rewritten to /api/avatar/{id} (keep query)
+// - http(s) or root-relative URLs: returned as-is
+// - otherwise: undefined (fallback)
+function normalizeAvatarUrl(url: string): string | undefined {
```
orangecoloured, line 37: This feels like an over-engineered way to get the avatar urls. Do we even have search query params in the avatar urls? I understand that always getting the url as `/avatar` without understanding whether it's an image or nothing is a bit confusing, but it's a simple logic. This update makes it look like we have some complex ways of getting the avatar urls, while in fact it stayed the same. I feel like this increases complexity of the code without any benefits. Maybe I'm missing something.

orangecoloured (changes requested):

apps/console/src/pages/api/avatar/[userId].ts
Comment (orangecoloured) [reply apps/console/src/pages/api/avatar/[userId].ts:1]: I mean that currently BE always returns `/v1beta/users/namespace/avatar`, not matter whether there is an avatar or not. Would be better to have it return either an empty string or an actual link to an image. This way the FE wouldn't have any confusion about the contents of the url.

orangecoloured (approved):

